PR #7, opened 2018-07-03
[ITBL-5301] - migrate new changes

tapashmajumder:
1. New initialization, email and userId storing in UserDefaults
2. New way of handling universal links.
3. More refactoring.


swift-sdk/IterableActionRunner.swift
```diff
@@ -44,35 +44,47 @@ class IterableActionRunner : NSObject, ActionRunnerProtocol {
...
         } else {
-            callCustomActionIfSpecified(action: action)
+            return callCustomActionIfSpecified(action: action, inContext: context)
         }
     }
     
-    private func open(url: URL, action: IterableAction) {
-        if urlDelegate?.handleIterableURL(url, fromAction: action) == true {
-            return
+    private func open(url: URL, inContext context: IterableActionContext) -> Bool {
```
davidtruong: Is there anything else that we need to consider in order to handle in-app notification link clicks?

swift-sdk/IterableAppIntegrationInternal.swift
```diff
@@ -205,4 +229,11 @@ struct IterableAppIntegrationInternal {
         return itbl
     }
 
+    private static func legacyDefaultActionFromPayload(userInfo: [AnyHashable : Any]) -> IterableAction? {
+        if let deeplinkUrl = userInfo[ITBConsts.Payload.deeplinkUrl] as? String {
```
davidtruong: I don't think ITBConsts.Payload.deeplinkUrl is set correctly, currently it references the constant: ITBL_PAYLOAD_DEEP_LINK_URL.

Shouldn't it be `URL` which is passed down from the backend?

davidtruong: Nevermind i see ITBL_PAYLOAD_DEEP_LINK_URL defined in the constants.

davidtruong: However, do you need to check case in-sensitivity?

swift-sdk/IterableAPI+Internal.swift
```diff
@@ -10,17 +10,111 @@ import Foundation
 
 /// Only things that are needed internally by IterableAPI
 extension IterableAPI {
+    /**
+     * Returns the push integration name for this app depending on the config options
+     * @return push integration name to use
+     */
+    var pushIntegrationName: String? {
+        if let pushIntegrationName = config.pushIntegrationName, let sandboxPushIntegrationName = config.sandboxPushIntegrationName {
```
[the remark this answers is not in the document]
tapashmajumder: Yes. In that case the method will return nil. That is expected and will be handled by the caller.

swift-sdk/IterableAppIntegrationInternal.swift
tapashmajumder: Right now we are not checking case insensitivity anywhere we are serializing/deserializing. Do we need to? I think we should be ok as long as this convention is followed by both iOS SDK and backend.

swift-sdk/IterableActionRunner.swift
tapashmajumder: I have not thought through in-app scenario here. If we need anything more for in-app we can add this to the IterableActionContext object. The point of having a context object there was so that this is extensible and we can add more "context" as we need.

Tests/swift-sdk-swift-tests/IterableNotificationResponseTests.swift
```diff
@@ -21,7 +21,7 @@ class IterableNotificationResponseTests: XCTestCase {
     override func setUp() {
         super.setUp()
         // Put setup code here. This method is called before the invocation of each test method in the class.
-        IterableAPI.clearInstance()
+        IterableAPI.clearSharedInstance()
```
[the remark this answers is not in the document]
tapashmajumder: This is only needed for tests. We have two choices either do IterableAPI.initialize again or call clearInstance(). I think it is probably best to remove clearSharedInstance() altogether and call initialize again.

swift-sdk/IterableAppIntegrationInternal.swift
davidtruong: The only reason I ask is that the documentation has `URL` as the key being passed down. https://support.iterable.com/hc/en-us/articles/115000315806-Setting-Up-iOS-Push-Notifications

vbabenkoru: Docs are wrong :)

davidtruong: Not any more they aren't :)

tapashmajumder: This seems wrong in the doc
```if (userInfouserInfo[deeplinkKey])```

davidtruong: Fixed the userInfo in the docs.

PR #8, opened 2018-07-09
[ITBL-5386]-add-ci-codecov

tapashmajumder:
Add ci and code coverage for Swift SDK. Make necessary changes to schemes, project files etc.


Tests/notification-extension-tests/NotificationExtensionTests.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+//  Created by Tapash Majumder on 7/6/18.
+//  Copyright © 2018 Iterable. All rights reserved.
+//
+
+import XCTest
+
+@testable import IterableAppExtensions
+
+class NotificationExtensionTests: XCTestCase {
```
vbabenkoru: Do we need this file?

notification-extension/TestFile.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  TestFile.swift
+//  notification-extension
+//
+//  Created by Victor Babenko on 7/6/18.
+//  Copyright © 2018 Iterable. All rights reserved.
+//
+import Foundation
+
+class TestFile {
```
vbabenkoru: It might be useful to add a short comment about why this file exists, to make sure we remember to delete it if we figure out a way to fix code coverage reports without this workaround.

Tests/notification-extension-tests/NotificationExtensionTests.swift
tapashmajumder: I don't think we need this. If we don't have this file then the coverage for project is reduced. I will rename this so that there is no confusion.

tapashmajumder:
Now we have ci and codecov. Finally.

PR #12, opened 2018-07-18
[ITBL-5586]-improvement-for-sample-apps

tapashmajumder:
- We now show an alert if apiKey is not set. 
- Separate out login/logout from api initialization to make it close to real life use.


sample-apps/objc-sample-app/objc-sample-app/LoginViewController.m
```diff
@@ -0,0 +1,52 @@
...
+- (IBAction)logInOutButtonTapped:(UIButton *)sender {
+    if (IterableAPI.email == nil) {
+        // login
+        if (self.emailAddressTextField.text != nil && self.emailAddressTextField.text.length > 0) {
+            IterableAPI.email = self.emailAddressTextField.text;
+            [[UIApplication sharedApplication] registerForRemoteNotifications];
+        }
+    } else {
+        // logout
+        [IterableAPI disableDeviceForCurrentUser];
```
davidtruong: What are your thoughts of automatically disabling the the device in our SDK when they change users? I think its a fairly common use case and people might forget to disable the old device. (Fine for a different PR)

tapashmajumder: Absolutely. We should move this to the SDK.

PR #16, opened 2018-07-26
[ITBL-5619] - Code coverage and refactoring

tapashmajumder:
Misc refactoring of code and add tests.


Tests/swift-sdk-swift-tests/IterableNotificationResponseTests.swift
```diff
@@ -90,14 +112,24 @@ class IterableNotificationResponseTests: XCTestCase {
...
+        promise.observe { (result) in
+            switch (result) {
+            case .error(let error):
+                XCTFail(error.localizedDescription)
+            case .value(let customActionName):
+                XCTAssertEqual(customActionName, "customAction")
+            }
+        }
+        
+        let appIntegration = IterableAppIntegrationInternal(tracker: pushTracker,
```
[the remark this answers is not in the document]
tapashmajumder: Yes makes sense. Although this is a change not introduced by this PR. I will fix it here.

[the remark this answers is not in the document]
vbabenkoru: I slightly prefer `Internal`/`internal`, but consistency is more important.

swift-sdk/NetworkHelper.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+//  Created by Tapash Majumder on 7/24/18.
+//  Copyright © 2018 Iterable. All rights reserved.
+//
+
+import Foundation
+
+typealias SendRequestValue = [AnyHashable : Any]
+
+struct SendRequestErrorType {
```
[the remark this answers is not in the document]
tapashmajumder: This is not NetworkError. We have to have both errorMessage and data for OnFailureCallback. Not sure why data is need though.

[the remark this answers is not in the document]
vbabenkoru: Looking at the struct name, I just thought it was a class specifying the type of the error it encountered (an enum).

swift-sdk/IterableAPI.swift
```diff
@@ -16,23 +16,23 @@ public final class IterableAPI : NSObject {
     /// You should call this method and not call the init method directly.
     /// - parameter apiKey: Iterable API Key.
     public static func initialize(apiKey: String) {
-        implementation = IterableAPIImplementation.initialize(apiKey: apiKey, launchOptions: nil, config:IterableConfig(), dateProvider: SystemDateProvider())
+        implementation = IterableAPIInternal.initialize(apiKey: apiKey)
```
vbabenkoru: Would it make sense to also rename this field from `implementation` to `internal`?

Tests/swift-sdk-swift-tests/IterableActionRunnerTests.swift
```diff
@@ -45,29 +47,56 @@ class IterableActionRunnerTests: XCTestCase {
...
     func testCustomAction() {
-        let customActionDelegate = MockCustomActionDelegate(returnValue: false)
-        let action = IterableAction.action(fromDictionary: ["type" : "customActionName"])!
-        let actionRunner = IterableActionRunner(urlDelegate: nil, customActionDelegate: customActionDelegate, urlOpener: MockUrlOpener())
-        _ = actionRunner.execute(action: action, from: .push)
+        let customActionName = "myCustomActionName"
+        let action = IterableAction.action(fromDictionary: ["type" : customActionName])!
+        let context = IterableActionContext(action: action, source: .push)
+        let customActionHandler: CustomActionHandler = {name in
+            XCTAssertEqual(name, customActionName)
```
vbabenkoru: Can we verify that this handler was called? If `execute` just retuns `true` without calling `customActionHandler`, the test will still pass.

vbabenkoru: Previously, we were using XCTest expectations for that. Not sure if there is a different API for Swift.

Tests/swift-sdk-swift-tests/IterableActionRunnerTests.swift
```diff
@@ -45,29 +47,56 @@ class IterableActionRunnerTests: XCTestCase {
...
+        XCTAssertTrue(handled)
     }
-    
+
+    func testCustomActionOverride() {
+        let customActionName = "myCustomActionName"
+        let action = IterableAction.action(fromDictionary: ["type" : customActionName])!
+        let context = IterableActionContext(action: action, source: .push)
+        let customActionHandler: CustomActionHandler = {name in
+            XCTAssertEqual(name, customActionName)
```
vbabenkoru: Same here

Tests/swift-sdk-swift-tests/IterableNotificationResponseTests.swift
```diff
@@ -50,14 +63,23 @@ class IterableNotificationResponseTests: XCTestCase {
 
         let response = MockNotificationResponse(userInfo: userInfo, actionIdentifier: UNNotificationDefaultActionIdentifier)
         let pushTracker = MockPushTracker()
-        let customActionDelegate = MockCustomActionDelegate(returnValue: true)
-        
-        let actionRunner = IterableActionRunner(urlDelegate: nil, customActionDelegate: customActionDelegate, urlOpener: MockUrlOpener())
-        let appIntegration = IterableAppIntegrationInternal(tracker: pushTracker, actionRunner: actionRunner, versionInfo: MockVersionInfo(version: 10))
+        let promise = Promise<String, Error>()
```
vbabenkoru: Why are promises necessary here? Wouldn't it be enough to just pass a mock `customActionHandler`?

swift-sdk/IterableActionRunner.swift
```diff
@@ -33,58 +29,120 @@ class AppUrlOpener : UrlOpenerProtocol {
...
+                return .customAction(action.type)
+            }
         }
-        guard let customActionDelegate = customActionDelegate else {
-            return false
+    }
+
+    // MARK: Helper Utility Functions
+    // converts from IterableURLDelegate to UrlHandler
+    static func urlHandler(fromUrlDelegate urlDelegate: IterableURLDelegate?, inContext context: IterableActionContext) -> UrlHandler {
```
vbabenkoru: I don't have a strong opinion on this, but do `UrlHandler`s offer any isolation vs. `IterableURLDelegate`s? It looks like `IterableActionContext` and `IterableURLDelegate` are still used in this file.

swift-sdk/IterableAppIntegrationInternal.swift
```diff
@@ -73,15 +73,21 @@ class SystemVersionInfo : VersionInfoProtocol {
...
     private let versionInfo: VersionInfoProtocol
+    private let contextToUrlHandler: ((IterableActionContext) -> UrlHandler)?
+    private let contextToCustomActionHandler: ((IterableActionContext) -> CustomActionHandler)?
+    private let urlOpener: UrlOpenerProtocol
 
     init(tracker: PushTrackerProtocol,
-         actionRunner: ActionRunnerProtocol,
-         versionInfo: VersionInfoProtocol) {
+         versionInfo: VersionInfoProtocol,
+         contextToUrlHandler: ((IterableActionContext) -> UrlHandler)?,
```
vbabenkoru: Also looks like `urlHandlers` add a bit of complexity here

swift-sdk/IterableActionRunner.swift
tapashmajumder: I think they do. These static methods can be moved to any other appropriate place later.

Tests/swift-sdk-swift-tests/IterableNotificationResponseTests.swift
tapashmajumder: Not really. I will change to expectations.

swift-sdk/IterableActionRunner.swift
tapashmajumder: I will move them to IterableUtil.

swift-sdk/NetworkHelper.swift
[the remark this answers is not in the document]
vbabenkoru: What does `Type` in the name mean?

Tests/swift-sdk-swift-tests/IterableActionRunnerTests.swift
```diff
@@ -27,12 +26,16 @@ class IterableActionRunnerTests: XCTestCase {
     
     func testUrlOpenAction() {
         let urlString = "https://example.com"
-        let urlDelegate = MockUrlDelegate(returnValue: false)
+        let action = IterableAction.action(fromDictionary: ["type" : "openUrl", "data" : urlString])!
+        let context = IterableActionContext(action: action, source: .push)
         let urlOpener = MockUrlOpener()
-        let actionRunner = IterableActionRunner(urlDelegate: urlDelegate, customActionDelegate: nil, urlOpener: urlOpener)
+
+        let handled = IterableActionRunner.execute(action: action,
```
vbabenkoru: I think this should also verify that urlHandler was called, similar to `testCustomAction` below.

swift-sdk/NetworkHelper.swift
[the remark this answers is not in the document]
tapashmajumder: Just signifies that it is not actually an 'Error' which is a Swift protocol. I suppose we could actually implement the Error protocol and call it SendRequestError instead of SendRequestErrorType.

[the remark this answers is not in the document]
vbabenkoru: Oh, you're right, I didn't think of `Error` as a separate protocol. But I think having `Type` suffix (as well as `Class`, `Struct`, etc.) as part of the actual name of the type isn't great since it doesn't add to the description the type itself and can be interpreted as something else (like a type/category of an error).
Maybe `NetworkErrorData`? (or implement `Error` if that isn't going to break things)

PR #18, opened 2018-08-08
[ITBL-5902] - Do now show in-app if one is already showing.

tapashmajumder:


swift-sdk/IterableAPIInternal.swift
```diff
@@ -616,10 +616,11 @@ import Foundation
...
             
             DispatchQueue.main.async {
-                IterableInAppManager.showIterableNotificationHTML(html, trackParams: notificationMetadata, callbackBlock: callbackBlock, backgroundAlpha: backgroundAlpha, padding: edgeInsets)
+                let opened = IterableInAppManager.showIterableNotificationHTML(html, trackParams: notificationMetadata, callbackBlock: callbackBlock, backgroundAlpha: backgroundAlpha, padding: edgeInsets)
+                if opened {
+                    self.inAppConsume(messageId)
+                }
             }
-
-            self.inAppConsume(messageId)
```
vbabenkoru: The ObjC implementation also consumes if 'href' isn't found (otherwise the invalid in-app may stay at the top of the queue until expiration), line 609 here.

tapashmajumder: Not sure I follow.

vbabenkoru: If you look at the check at line 608, it prints an error, but still tries to display the invalid in-app message (it shouldn't) and consumes it (it should). After logging the error, it should do `self.inAppConsume(messageId)` and `return` if the message is invalid.

swift-sdk/IterableAPIInternal.swift
```diff
@@ -605,8 +605,10 @@ import Foundation
             guard let html = message[ITERABLE_IN_APP_HTML] as? String else {
                 return
             }
-            if html.range(of: ITERABLE_IN_APP_HREF, options: [.caseInsensitive]) == nil {
+            guard html.range(of: ITERABLE_IN_APP_HREF, options: [.caseInsensitive]) != nil else {
```
vbabenkoru: 👍

PR #20, opened 2018-08-15
[ITBL-5848] - disable/re-register device token when email or userId changes

tapashmajumder:


swift-sdk/IterableAppIntegrationInternal.swift
```diff
@@ -9,6 +9,40 @@
...
+            UNUserNotificationCenter.current().getNotificationSettings { settings in
+                if settings.authorizationStatus == .authorized {
+                    result.resolve(with: true)
+                } else {
+                    result.resolve(with: false)
+                }
+            }
+        } else {
+            // Fallback on earlier versions
+            if UIApplication.shared.isRegisteredForRemoteNotifications {
```
vbabenkoru: Use `currentUserNotificationSettings` instead, `isRegisteredForRemoteNotifications` only indicates whether we got the token or not.

swift-sdk/IterableAPIInternal.swift
```diff
@@ -849,16 +850,51 @@ import Foundation
         return characterSet
     } ()
     
+    private func isEitherUserIdOrEmailSet() -> Bool {
+        return IterableUtil.isNotNullOrEmpty(string: _email) || IterableUtil.isNotNullOrEmpty(string: _userId)
+    }
+    
+    private func disablePrevious() {
```
vbabenkoru: I'm not sure about this name.. it doesn't tell what exactly it disables.

swift-sdk/IterableAppIntegrationInternal.swift
```diff
@@ -9,6 +9,40 @@
 import Foundation
 import UserNotifications
 
+// Returns whether notifications are enabled
+protocol NotificationStateProviderProtocol {
+    var notificationsEnabled : Promise<Bool, Error> {get}
+    func registerForRemoteNotification()
```
vbabenkoru: `registerForRemoteNotifications`?

vbabenkoru:
Looks like CI tests are timing out on `testDisableDeviceForCurrentUser`

swift-sdk/IterableAPIInternal.swift
```diff
@@ -854,7 +854,7 @@ import UserNotifications
         return IterableUtil.isNotNullOrEmpty(string: _email) || IterableUtil.isNotNullOrEmpty(string: _userId)
     }
     
-    private func disablePrevious() {
+    private func disableDevicePreviousUser() {
```
vbabenkoru: `disableDeviceForPreviousUser`

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
```diff
@@ -0,0 +1,207 @@
...
+        let config = IterableConfig()
+        config.pushIntegrationName = "my-push-integration"
+        config.autoPushRegistration = true
+        
+        let notificationStateProvider = MockNotificationStateProvider(enabled: true)
+        var registerForRemoteNotificationsCount = 0
+        notificationStateProvider.callback = {
+            registerForRemoteNotificationsCount += 1
+            
+            if registerForRemoteNotificationsCount == 2 {
```
vbabenkoru: Can we check `registerForRemoteNotificationsCount` after `wait(for:` at the end of the test instead? That way we won't need this logic in the callback (and the callback itself).

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
```diff
@@ -0,0 +1,207 @@
...
+                expectation2.fulfill()
+            }
+        }
+        
+        IterableAPI.initialize(apiKey: IterableAutoRegistrationTests.apiKey, config:config, networkSession: networkSession, notificationStateProvider: notificationStateProvider)
+        IterableAPI.email = "user1@example.com"
+        let token = "zeeToken".data(using: .utf8)!
+        networkSession.callback = {(_, _, _) in
```
vbabenkoru: I think it's generally better to store requests in an array in the mock network session and then verify their content, that way the control flow would be easier to understand.

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
```diff
@@ -0,0 +1,207 @@
...
+            networkSession.callback = {(_, _, _)in
+                // Second callback should not happen
+                XCTFail("Should not call disable")
+            }
+            // set same value
+            IterableAPI.email = email
+            
+            // wait for 1 second to make sure that registerForRemoteNotifications is not called
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1, execute: {
+                XCTAssertEqual(registerForRemoteNotificationsCount, 1) // should be called only once (when setting user1@example.com).
```
vbabenkoru: What do you think about modifying the mocks to take expectations and then fulfill them in the callback? Like, add an expectation parameter to MockNotificationStateProvider, and then you'd be able to use `expectedFulfillmentCount` to verify the call count.

tapashmajumder: Brilliant!

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
tapashmajumder: I like the idea. I will do that when I work on codecov part 3.

swift-sdk/IterableAPIInternal.swift
```diff
@@ -872,6 +908,10 @@ import Foundation
         // get email and userId from UserDefaults if present
         retrieveEmailAndUserId()
         
+        if config.autoPushRegistration == true && isEitherUserIdOrEmailSet() {
+            notificationStateProvider.registerForRemoteNotifications()
```
vbabenkoru: This may be called on a background thread, maybe `SystemNotificationStateProvider` should wrap its call in `DispatchQueue.main.async` so we don't have to do that in the caller?

swift-sdk/IterableAppIntegrationInternal.swift
```diff
@@ -9,6 +9,40 @@
...
+            } else {
+                result.resolve(with: false)
+            }
+        }
+        
+        return result
+    }
+
+    func registerForRemoteNotifications() {
+        UIApplication.shared.registerForRemoteNotifications()
```
vbabenkoru: What do you think about wrapping this in `DispatchQueue.main.async` and remove it from `enableDeviceForCurrentUser`?

swift-sdk/IterableConfig.swift
```diff
@@ -64,4 +64,8 @@ public class IterableConfig : NSObject {
 
     /// How to handle IterableActions which are other than 'openUrl'
     public weak var customActionDelegate: IterableCustomActionDelegate?
+    
+    /// When set to true, IterableSDK will automatically register and deregister for
```
vbabenkoru: Not sure, but looks like `for` can be removed

Tests/swift-sdk-swift-tests/Mocks.swift
```diff
@@ -208,14 +210,17 @@ class MockNotificationStateProvider : NotificationStateProviderProtocol {
     var callback: (() -> Void)? = nil
 
     func registerForRemoteNotifications() {
+        expectation?.fulfill()
         callback?()
```
vbabenkoru: Do we stil need the callback?

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
```diff
@@ -173,35 +148,31 @@ class IterableAutoRegistrationTests: XCTestCase {
...
         
         let networkSession = MockNetworkSession(statusCode: 200)
         let config = IterableConfig()
         config.pushIntegrationName = "my-push-integration"
         config.autoPushRegistration = true
-        let notificationStateProvider = MockNotificationStateProvider(enabled: true) // Notifications are on.
+        let notificationStateProvider = MockNotificationStateProvider(enabled: true, expectation: expectation2)
         notificationStateProvider.callback = {
-            expectation.fulfill()
+            expectation1.fulfill()
```
vbabenkoru: Why is it using both the expectation and the callback?

Tests/swift-sdk-swift-tests/IterableAutoRegistrationTests.swift
```diff
@@ -67,21 +60,18 @@ class IterableAutoRegistrationTests: XCTestCase {
     }
     
     func testDoNotCallDisableAndEnableWhenSameValue() {
-        let expectation = XCTestExpectation(description: "testDoNotCallDisableAndEnableWhenSameValue")
+        let expectation1 = expectation(description: "registerForRemoteNotifications")
+        expectation1.expectedFulfillmentCount = 1
```
vbabenkoru: 1 is the default, do we need this?

tapashmajumder: yes explicit setting is not needed here.

Tests/swift-sdk-swift-tests/Mocks.swift
tapashmajumder: We can remove it.

PR #21, opened 2018-09-04
[itbl-5833] - Deferred Deeplinking

tapashmajumder:
Add deferred deep linking to SDK.
  1. Add config option for DDL (deferred deep linking).
  2. Send deviceInfo to server and do DDL on first launch.
  3. Little bit of refactoring of how to do local storage.

### DIFF
--- a/swift-sdk/DeviceInfo.swift
+++ b/swift-sdk/DeviceInfo.swift
@@ -34,7 +34,9 @@ struct DeviceInfo : Codable {
     private static func createDeviceFp() -> DeviceFp {
         let screen = UIScreen.main
         let device = UIDevice.current
-        // secondsFromGMT = time - GMT, timeZoneOffset = GMT - time
+        // iOS TimeZone stores secondsFromGMT which is the difference between the local time and GMT in seconds.
+        // We are supposed to return 'timeZoneOffset', which is difference between GMT and local time in minutes.
+        // Therefore, the conversion from secondsFomGMT to timezoneOffsetMinutes is as follows.
         let secondsFromGMT = TimeZone.current.secondsFromGMT()
         let timezoneOffsetMinutes = (-1.0 * Float(secondsFromGMT) / 60.0)
         return DeviceFp(iosDeviceType: getModelName(),
@@ -70,6 +72,9 @@ struct DeviceInfo : Codable {
         case "iPhone9,1", "iPhone9,3":                  return "iPhone 7"
         case "iPhone9,2", "iPhone9,4":                  return "iPhone 7 Plus"
         case "iPhone8,4":                               return "iPhone SE"
+        case "iPhone10,1", "iPhone10,4":                return "iPhone 8"
+        case "iPhone10,2", "iPhone10,5":                return "iPhone 8 Plus"
+        case "iPhone10,3", "iPhone10,6":                return "iPhone X"
         case "iPad2,1", "iPad2,2", "iPad2,3", "iPad2,4":return "iPad 2"
         case "iPad3,1", "iPad3,2", "iPad3,3":           return "iPad 3"
         case "iPad3,4", "iPad3,5", "iPad3,6":           return "iPad 4"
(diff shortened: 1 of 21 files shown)


swift-sdk/ITBConsts.swift
```diff
@@ -42,4 +32,7 @@ import Foundation
         public static let inputPlaceholder = ITBL_BUTTON_INPUT_PLACEHOLDER
         public static let action = ITBL_BUTTON_ACTION
     }
+
+    private static let apiHostname = "https://api.iterable.com"
```
davidtruong: Any reason why this is separated from the endpoints?
Are you thinking of reusing these constants?

swift-sdk/IterableAPIInternal+Extension.swift
```diff
@@ -13,7 +13,7 @@ extension IterableAPIInternal {
      * Returns the push integration name for this app depending on the config options
      * @return push integration name to use
      */
-    var pushIntegrationName: String? {
+    var pushIntegrationName: String? {//!!TQM: Take another look.
```
davidtruong: Remove comment if not needed

swift-sdk/IterableConstants.h
```diff
@@ -62,6 +62,7 @@ extern NSString *const ITBL_DEVICE_USER_INTERFACE;
 #define ENDPOINT_UPDATE_USER @"users/update"
 #define ENDPOINT_UPDATE_EMAIL @"users/updateEmail"
 #define ENDPOINT_UPDATE_SUBSCRIPTIONS @"users/updateSubscriptions"
+#define ENDPOINT_DDL_MATCH @"a/matchFp" //DDL = Deferred Deep Linking
```
davidtruong: What exactly is matchFp?

swift-sdk/DeviceInfo.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+    
+    static func createDeviceInfo() -> DeviceInfo {
+        return DeviceInfo(deviceFp: createDeviceFp())
+    }
+    
+    private static func createDeviceFp() -> DeviceFp {
+        let screen = UIScreen.main
+        let device = UIDevice.current
+        // secondsFromGMT = time - GMT, timeZoneOffset = GMT - time
```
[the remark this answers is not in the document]
tapashmajumder: I think this comment here helps. I was getting a little confused by how things work.

swift-sdk/DeviceInfo.swift
```diff
@@ -0,0 +1,89 @@
...
+    private static func getModelName() -> String {
+        var systemInfo = utsname()
+        uname(&systemInfo)
+        let machineMirror = Mirror(reflecting: systemInfo.machine)
+        let identifier = machineMirror.children.reduce("") { identifier, element in
+            guard let value = element.value as? Int8, value != 0 else { return identifier }
+            return identifier + String(UnicodeScalar(UInt8(value)))
+        }
+        
+        switch identifier {
```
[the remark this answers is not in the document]
tapashmajumder: For the devices that we can map, we return the known names such as 'iPhone 7 Plus' as  you have noticed. For the ones that are not known we will still get the identifier "iPhone10,3" or "iPhone11,2"... something like that. 

Right now how the matching logic works is that... .the device sends all information that it has and Javascript sends all information that it has. Then the matching engine tries to match the two. The model names are not important for the matching engine at the moment. It will look at the screen size, os version, screen width, height, ipAddress time etc. The model information is not used right but but it is there and may be used in future matching strategies. 

I can add iPhone X details in here. The caveat is that we won't have iPhone XS when it comes out.

swift-sdk/ITBConsts.swift
tapashmajumder: I find it easier to switch to 'localhost:9000' etc when testing. I just have to change the hostname.

swift-sdk/IterableConstants.h
tapashmajumder: This is the endpoint that will be called for the matching deferred deeplinking fingerprints.(fp = fingerprinting)

swift-sdk/DeviceInfo.swift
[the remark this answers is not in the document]
tapashmajumder: I am beginning to think passing this information does not provide a lot of value at the moment so perhaps I can remove this. We can add this back later if needed. That will lead to less confusion. We can just use `UserInterfaceIdiom` which is provided by the system.

swift-sdk/IterableConstants.h
davidtruong: Got it, we still need to implement that endpoint.

tapashmajumder: Yes. I have already implemented that endpoint in the 'Iterable' repo but I haven't merged my PR yet.

PR #23, opened 2018-09-07
[itbl-5996] - add additional fields

tapashmajumder:
Add additional fields. Also refactor IterableAPIInternal file.


swift-sdk/ITBConsts.swift
```diff
@@ -32,6 +32,20 @@ import Foundation
...
+        public static let identifierForVendor = "identifierForVendor"
+        public static let model = "model"
+        public static let systemName = "systemName"
+        public static let systemVersion = "systemVersion"
+        public static let userInterfaceIdiom = "userInterfaceIdiom"
+        public static let deviceId = "deviceId"
+        public static let appPackageName = "appPackageName"
+        public static let appVersion = "appVersion"
+        public static let appBuild = "appBuild"
+        public static let sdkVersion = "sdkVersion"
```
vbabenkoru: Sorry the original ticket description/spec was incorrect, turns out there is a field called `sdkVersion`, and it's used for Android platform version (API level). I've renamed our field to iterableSdkVersion in the Android implementation.

tapashmajumder: Ok so rename sdkVersion => iterableSdkVersion?

vbabenkoru: Yup

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -143,59 +103,94 @@ import UserNotifications
...
+            ]
+        } else if let userId = userId {
+            args = [
+                ITBL_KEY_USER_ID: userId,
+                ITBL_KEY_DEVICE: deviceDictionary
+            ]
+        } else {
+            ITBError("Either email or userId is required.")
+            assertionFailure("either email or userId should be set")
+            args = [
```
davidtruong: Is setting the args necessary if you are asserting an error?

davidtruong: Also should we not attempt to send a request when the email/userId is not set.

swift-sdk/Internal/UserDefaultsLocalStorage.swift
```diff
@@ -141,7 +157,7 @@ struct UserDefaultsLocalStorage : LocalStorageProtocol {
         
         let envelope = Envelope(payload: data, expiration: expiration)
         let encodedEnvelope = try JSONEncoder().encode(envelope)
-        UserDefaults.standard.set(encodedEnvelope, forKey: key.value)
+        UserDefaults.standard.set (encodedEnvelope, forKey: key.value)
```
davidtruong: minor style: extra space

swift-sdk/Internal/IterableUtil.swift
```diff
@@ -29,6 +29,10 @@ import os
         return !isNullOrEmpty(string: string)
     }
     
+    static func generateUUID() -> String {
+        return UUID().uuidString.replacingOccurrences(of: "-", with: "")
```
davidtruong: Any particular reason you wanted to remove the "-" in the UUID?
What if we need to obtain hashValue of the UUID?

tapashmajumder: It is just a matter of preference. I don't think it matters as long as the values are unique. I will change it so that we don't replace the '-'s anymore.

Since the generated UUID is a string. The hashValue will have the same semantics as String so we should be fine.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: Please note that code will never enter line 165 above because of the guard statement at the beginning of the function. But I agree that it is not very clear. I will change the code so that it is more clear.

swift-sdk/Internal/IterableUtil.swift
davidtruong: I was thinking if we used the built in HashValue function it probably uses the "-" as part of the string to hash. UUIDs are pretty standard with the "-", so it might be simpler to leave it in.

vbabenkoru:
I think we should have separate PRs (maybe without a jira ticket) for refactoring changes. Having feature changes and medium to large scale refactors separated into different branches/PRs would make them easier to review.

Tests/swift-sdk-swift-tests/LocalStorageTests.swift
```diff
@@ -92,4 +92,20 @@ class LocalStorageTests: XCTestCase {
         let fromLocalStorage2:[AnyHashable : Any]? = localStorage.payload
         XCTAssertNil(fromLocalStorage2)
     }
+    
+    func testDeviceId() {
+        let mockDateProvider = MockDateProvider()
+        var localStorage: LocalStorageProtocol = UserDefaultsLocalStorage(dateProvider: mockDateProvider)
+        let deviceId = UUID().uuidString.replacingOccurrences(of: "-", with: "")
```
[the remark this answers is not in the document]
tapashmajumder: This will not fail because what we are testing here is only saving/retrieval of the value. We could have set any valid string here.  This is just testing whether the value set is the value saved in the storage.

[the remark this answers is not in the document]
tapashmajumder: But I think you are right. Removing 'replacingOccurrences' is less confusing and more standard so I will remove it.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -0,0 +1,819 @@
...
+                                             context: context,
+                                             urlHandler: IterableUtil.urlHandler(fromUrlDelegate: self.urlDelegate, inContext: context),
+                                             urlOpener: AppUrlOpener())
+            }
+        }
+        
+        localStorage.ddlChecked = true
+    }
+    
+    private func updateDeviceId() {
```
vbabenkoru: What do you think about initializing deviceId lazily by moving this into the getter method?

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -0,0 +1,819 @@
...
+    
+    private func updateSDKVersion() {
+        if let lastVersion = localStorage.sdkVersion, lastVersion != IterableAPI.sdkVersion {
+            performUpgrade(lastVersion: lastVersion, newVersion: IterableAPI.sdkVersion)
+        } else {
+            localStorage.sdkVersion = IterableAPI.sdkVersion
+        }
+    }
+    
+    private func performUpgrade(lastVersion: String, newVersion: String) {
```
vbabenkoru: `String` isn't very useful when determining which upgrades to perform; I think transforming the version (1.2.3) into an int (1002003) would let us specify upgrade paths. That way, you can do `if lastVersion < 1002003` to figure out whether you need a specific upgrade. Alternatively, there could be a method like `shouldUpgradeFrom(oldVersion: String, newVersion: String)` that would do String -> int conversions and comparison internally.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: Not sure. Do we gain much be creating a separate property on IterableSDK. I don't think it matters much either way.

tapashmajumder: On the other hand it is probably a little cleaner.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: We can add those later. I would rather design this properly when we actually do upgrades. We have the plumbing here

vbabenkoru: Yeah that's not necessary for now.

tapashmajumder:
I have changed how we update device id. Please take a look again. @vbabenkoru

vbabenkoru:
If we at some point need to access `deviceId` elsewhere, we'll need to move this or make sure we don't fetch deviceId before calling `register`. I'd make this a field, but that's not a strong opinion.

PR #28, opened 2018-10-22
Feature/itbl 6229 call create user for user

davidtruong:
Attempts to create a user profile when the userId is set.

### DIFF
--- a/swift-sdk/Internal/IterableAPIInternal.swift
+++ b/swift-sdk/Internal/IterableAPIInternal.swift
@@ -236,7 +236,7 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
         }
     }
 
-    func createUser(withUserId userId: String, onSuccess: OnSuccessHandler?, onFailure: OnFailureHandler?) {
+    private func createUser(withUserId userId: String, onSuccess: OnSuccessHandler?, onFailure: OnFailureHandler?) {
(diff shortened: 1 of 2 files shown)


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -53,6 +53,7 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
             _userId = newValue
             _email = nil
             storeEmailAndUserId()
+            createUserForUserId(onSuccess: IterableAPIInternal.defaultOnSucess(identifier: "createUserForUserId"), onFailure: IterableAPIInternal.defaultOnFailure(identifier: "createUserForUserId"))
```
tapashmajumder: We should probably wait for the call to createUserForUserId to finish before calling enableDevice. That way we know that the user exists before enabling the device. Also adding a unit test would be good.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -221,6 +236,15 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
         }
     }
 
+    func createUser(withUserId userId: String, onSuccess: OnSuccessHandler?, onFailure: OnFailureHandler?) {
```
vbabenkoru: Let's make it private

PR #29, opened 2018-10-30
[ITBL-6407] - Promise and Future Refactor

tapashmajumder:
Add concept of map and flatMap to Futures.


swift-sdk/Internal/Promise.swift
```diff
@@ -0,0 +1,118 @@
...
+    case value(Value)
+    case error(ErrorType)
+}
+
+// This has only two public methods
+// either there is a success with result
+// or there is a failure with error
+// There is no way to set value a result in this class.
+class Future<Value, ErrorType> {
+    var onSuccess: ((Value) -> Void)? = nil {
```
vbabenkoru: Isn't this usually a function that returns another `Future` so you can chain these?
```
future.flatMap({...}).onSuccess({...}).onFailure({...})
```

PR #30, opened 2018-11-05
[MOB-1] - Use Future Internally

tapashmajumder:
Renamed onFailure to onError for Future.
Refactor spawnInApp method.

### DIFF
--- a/swift-sdk/Internal/IterableAPIInternal.swift
+++ b/swift-sdk/Internal/IterableAPIInternal.swift
@@ -399,27 +399,30 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
         }
     }
     
-    func spawn(inAppNotification callbackBlock:ITEActionBlock?) {
-        getInAppMessages(1).flatMap { IterableInAppManager.handleInApp(withPayload: $0, callbackBlock: callbackBlock) }.onSuccess {
-            switch $0 {
-            case .success(opened: let opened, messageId: let messageId):
-                if opened {
-                    self.inAppConsume(messageId)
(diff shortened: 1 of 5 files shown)


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -399,56 +399,34 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
...
+            .flatMap { IterableInAppManager.showInApp(parseResult: $0, callbackBlock: callbackBlock)}
+            .onSuccess {
+                switch $0 {
+                case .success(opened: let opened, messageId: let messageId):
+                    if opened {
+                        self.inAppConsume(messageId)
+                    }
+                case .failure(reason: let reason, messageId: let messageId):
+                    if let messageId = messageId {
+                        self.inAppConsume(messageId)
```
vbabenkoru: Hmm looks like it wasn't consuming the in-app if it was not shown, but now it'll consume it even when it fails

vbabenkoru: Oh, it looks like it will return success with opened = false if it was shown, but failure if it could not be parsed.. Was it doing the same before the change?

tapash-majumder: Actually it will return success with opened = true if it was shown, false otherwise (if there is already an in-app message showing etc.).
If there is an error, it will return with messageId if one is available. 
This was the behavior before the change.

davidtruong: Correct that was the original behavior. If the content of the html is invalid we don't attempt to show it and we remove it from the queue. IE: if the html doesn't have a href button then don't show it otherwise there is no way to close the dialog.

PR #31, opened 2018-11-16
[mob-3] - In App Arch Change

tapashmajumder:
Make inApp based on server push rather than client pull. We now have callbacks in the SDK to inform of new inApps arriving etc.


swift-sdk/IterableInApp.swift
```diff
@@ -0,0 +1,125 @@
...
+@objc
+public protocol IterableInAppManagerProtocol {
+    /// - returns: A list of all messages
+    @objc(getMessages) func getMessages() -> [IterableInAppMessage]
+
+    /// - parameter message: The message to show.
+    /// - parameter consume: Set to true to consume the event from the server queue if the message is shown. This should be default.
+    /// - parameter callback: block of code to execute once the user clicks on a link or button in the inApp notification.
+    ///   Note that this callback is called in addition to calling `IterableCustomActionDelegate` or `IterableUrlDelegate` on the button action.
+    @objc(showContent:consume:callbackBlock:) func show(message: IterableInAppMessage, consume: Bool, callback:ITEActionBlock?)
```
vbabenkoru: `showContent` should be `showMessage` here I guess; there are 1 or 2 places elsewhere where `show(contentL:` or `showContent` is mentioned.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -0,0 +1,155 @@
...
+            message.skipped = true
+            messagesMap.updateValue(message, forKey: message.messageId)
+        } else {
+            // Should never happen
+            ITBError("Did not find message")
+        }
+    }
+
+    // Adds new messages to map and returns the new messages
+    private func mergeAndGetNewMessages(messages: [IterableInAppMessage]) -> [IterableInAppMessage] {
```
vbabenkoru: It should also remove messages that are no longer in the queue.

PR #35, opened 2018-12-20
[MOB-75] - Further improvements to async in-app messages.

tapashmajumder:


swift-sdk/IterableConfig.swift
```diff
@@ -140,4 +131,7 @@ public class IterableConfig : NSObject {
     /// By default, every single inApp will be shown as soon as it is available.
     /// If more than 1 inApp is available, we show the first.
     public var inAppDelegate: IterableInAppDelegate = DefaultInAppDelegate()
+
+    /// How many seconds to wait before showing the next inApp, if there are more than one present
+    public var newInAppMessageCallbackIntervalInSeconds: Double = 30.0
```
vbabenkoru: `inAppDisplayInterval`?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -18,46 +26,121 @@ class InAppManager : IterableInAppManagerProtocol {
...
+    
+    deinit {
+        ITBInfo()
+        notificationCenter.removeObserver(self)
     }
     
     func getMessages() -> [IterableInAppMessage] {
         ITBInfo()
-        return Array(messagesMap.values)
+        return Array(messagesMap.values.filter { $0.consumed == false })
```
vbabenkoru: This list should keep the order as it was in the remote queue (and new messages should be added to the end of the queue).

swift-sdk/Internal/InAppManager.swift
```diff
@@ -18,46 +26,121 @@ class InAppManager : IterableInAppManagerProtocol {
...
-            } else {
-                message.skipped = true
-                self.messagesMap.updateValue(message, forKey: message.messageId)
-            }
+        // set processed, if not already set
+        message.processed = true
+
+        let showed = displayer.showInApp(message: message, callback: clickCallback)
+        if showed {
+            lastShowedTime = dateProvider.currentDate
```
vbabenkoru: `lastShowedTime` should be set when the in-app is dismissed, so we count the interval as the time between in-app dismiss and next in-app show.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -84,75 +168,119 @@ class InAppManager : IterableInAppManagerProtocol {
...
-    // Adds new messages to map and returns the new messages
-    private func mergeAndGetNewMessages(messages: [IterableInAppMessage]) -> [IterableInAppMessage] {
-        return messages.reduce(into: [IterableInAppMessage]()) { (result, message) in
-            if !messagesMap.contains(where: { $0.key == message.messageId}) {
-                messagesMap[message.messageId] = message
-                result.append(message)
+        for message in messagesMap.values.filter({ $0.processed == false }) {
+            ITBDebug("campaignId: \(message.campaignId)")
+            message.processed = true
+            if inAppDelegate.onNew(message: message) == .show {
```
vbabenkoru: Since we're going all the way to `showInApp` before we know if it can be shown, it may be possible that the app will receive a new in-app message while another is on the screen, call `onNew` and then fail to display the message in the displayer. Is there a way to add a check before that?

tapashmajumder: If showInApp fails for newly arriving message due to a message already being shown then it will retry after retryInterval.

tapashmajumder: Basically showOneMessage keeps retrying until it succeeds. Hmm... although I can see a problem with this if at the end of retryInterval the app goes to background.

vbabenkoru: Oh yeah you're right. Then the issue is that if the app goes to background within 30s, we'll never show this message because it was marked as processed but not shown. Even if the app doesn't go to background, the in-app message may be shown at some later point, and the app won't know when. onNew should probably be called right before the SDK is going to show the message (and can do that) so that the app can decide if now is a good time to show it.

tapashmajumder: Makes sense. For that we will need to add a method to check for 'canShowMessage()' method on the InAppDisplayer. I will do that.

PR #36, opened 2019-01-09
[MOB-80] - Persist inapt.

tapashmajumder:


swift-sdk/Internal/InAppHelper.swift
```diff
@@ -301,15 +301,15 @@ struct InAppHelper {
...
+    /// This is a struct equivalent of IterableInAppMessage class
     private struct InAppDetails {
         let channelName: String
         let messageId: String
         let campaignId: String
         let edgeInsets: UIEdgeInsets
         let backgroundAlpha: Double
         let html: String
-        let extraInfo: [AnyHashable : Any]?
+        let extraInfo: [String : String]?
```
vbabenkoru: This user-defined JSON payload can be nested, and if we serialize it as key-value, it might not be the same as the input when deserialized from a file.

swift-sdk/Internal/InAppPersistence.swift
```diff
@@ -0,0 +1,175 @@
...
+
+protocol InAppPersistenceProtocol {
+    func getMessages() -> [IterableInAppMessage]
+    func persist(_ messages: [IterableInAppMessage])
+    func clear()
+}
+
+
+class FilePersister : InAppPersistenceProtocol {
+    init(filename: String = "itbl", ext: String = "json") {
```
vbabenkoru: We may need to store multiple files for other things in the future, what do you think about adding inapp to the name? Like `itbl_inapp.json` or similar.

swift-sdk/Internal/InAppHelper.swift
```diff
@@ -384,7 +384,7 @@ struct InAppHelper {
     }
     
     private static func parseExtraInfo(fromContent content: [AnyHashable : Any]) -> [AnyHashable : Any]? {
-        return content[.ITBL_IN_APP_PAYLOAD] as? [AnyHashable : Any]
+        return content[.ITBL_IN_APP_PAYLOAD] as? [String : String]
```
vbabenkoru: Revert to `[AnyHashable : Any]`?

PR #40, opened 2019-01-25
[MOB-131] - Sync on Move to Foreground Etc.

tapashmajumder:


swift-sdk/Internal/InAppManager.swift
```diff
@@ -215,23 +243,23 @@ extension InAppManager : InAppSynchronizerDelegate {
...
                 }
             } else {
-                self.processMessages()
+                self.processNextMessage()
             }
         }
     }
     
-    private func processMessages() {
+    private func processNextMessage() {
```
vbabenkoru: Why is is `nextMessage` now?

tapashmajumder:
I think the name makes it clearer. This method processes the next ‘processable’ message.

swift-sdk/Internal/InAppManager.swift
tapashmajumder: It is the basically the same method just the name is changed. I think it makes the intention clearer. It process the next 'processable' message. It will not process all messages just the first processable message.

PR #41, opened 2019-01-25
[MOB-136] - InApp Trigger Type Support

tapashmajumder:


vbabenkoru:
I think payload spec will be:
```
"trigger": {
    "type":"immediate"
}
```
It looks like this implementation is expecting this:
```

**Comment by tapashmajumder** [comment]
Right. I forgot about that. You may want to add this to Jira description.

File: swift-sdk/Internal/InAppHelper.swift [hunk 1]
```diff
@@ -375,12 +377,33 @@ struct InAppHelper {
...
+        
+        switch triggerTypeString.lowercased() {
+        case String(describing: IterableInAppTriggerType.immediate).lowercased():
+            return .immediate
+        case String(describing: IterableInAppTriggerType.event).lowercased():
+            return .event
+        case String(describing: IterableInAppTriggerType.never).lowercased():
+            return .never
+        default:
+            return .immediate
```
vbabenkoru: I think it's better to default to `.never`

tapashmajumder: Agreed. We should open a Jira for this.

PR #50, opened 2019-02-04
[MOB-177] - Add deferred deep linking documentation

tapashmajumder:


README.md
```diff
@@ -248,16 +248,24 @@ func application(_ application: UIApplication, continue userActivity: NSUserActi
 	return [IterableAPI handleUniversalLink:userActivity.webpageURL];
 }
 ```
+
+### Deferred deep linking
+
+If an app user does not have your app installed, deep linking will not work. The user will need to install your app first. After installing your app, the user will be taken to a specific location within your app when they launch it for the first time. See definition of deferred deep linking [here](#https://en.wikipedia.org/wiki/Deferred_deep_linking).
```
bradumbaugh: - I'm not quite sure about this part—maybe flesh out the sequence of events a bit more? So, a user clicks on a link that would take you to your app, but you don't have it installed... so it directs you to the app store, installs it, and when you open it, it takes you to the desired content? Is that how it works? I think it's a little confusing to say "deep linking will not work," because with this feature deep linking _will_ work even if you haven't installed the app yet, right?

README.md
```diff
@@ -248,16 +248,24 @@ func application(_ application: UIApplication, continue userActivity: NSUserActi
...
+
+#### Enabling deferred deep linking
+
+Set `IterableConfig.checkForDeferredDeeplink = true` to enable deferred deep linking for IterableSDK.
```
bradumbaugh: - Is "IterableSDK" the name of the SDK? It reads a little awkwardly, but if that's the title, then 👍

README.md
tapashmajumder: You are right about how ddl works. Let me see if I can rephrase it better.

README.md
tapashmajumder: In cocoapods it is called 'Iterable-iOS-SDK'. In your code the name of the module is 'IterableSDK'.  That is what you use to import our SDK in code. I think we should call it 'IterableSDK'.

README.md
```diff
@@ -248,16 +248,43 @@ func application(_ application: UIApplication, continue userActivity: NSUserActi
 	return [IterableAPI handleUniversalLink:userActivity.webpageURL];
 }
 ```
+
+### Deferred deep linking
+
+Deferred deep linking comes into play when a user of your app does not have your app installed in their iPhone/iPad. See definition of deferred deep linking [here](#https://en.wikipedia.org/wiki/Deferred_deep_linking). If you enable deferred deeplinking in IterableSDK, when the user clicks on a deep link in an email, the user is first sent to the App Store. After installing your app from the App Store, the user is taken to a specific location within your app when they launch it for the first time.
```
bradumbaugh:
```suggestion
[Deferred deep linking](https://en.wikipedia.org/wiki/Deferred_deep_linking) allows a user who does not have an app installed to:

 - Click on a link related to that app.
 - Install the app from the App Store.
 - Open the app and immediately see the content referenced by the link.
 
As the name implies, the deep link is _deferred_ until the app has been installed. 

After tapping a deep link in an email from an Iterable campaign, users without the associated app will be directed to the App Store to install it. If the app uses the IterableSDK and has deep linking enabled, the content associated with the deep link will load on first launch.
```

bradumbaugh: @tapashmajumder Took a crack at this—this was a surprisingly confusing thing to write. Let me know what you think!

README.md
```diff
@@ -248,16 +248,43 @@ func application(_ application: UIApplication, continue userActivity: NSUserActi
...
+
+Set `IterableConfig.checkForDeferredDeeplink = true` to enable deferred deep linking for IterableSDK.
 	
-#### In-app messages
+### In-app messages
+
+If you are already using in-app messages with IterableSDK, please check out the [migration section](#Migrating-in-app-messages-from-the-previous-version-of-the-SDK). 
 
-If you are already using in-app messages with IterableSDK, please check out the [migration section](#Migrating-in-app-messages-from-previous-version-of-SDK).
+In-app messages are handled via silent push messages from the server. When your application receives a silent push, you need to call IterableSDK in your AppDelegate as in the following code:
```
bradumbaugh: - You don't _need_ to do this, right? If you don't, the SDK will default to syncing the queue and showing the newest message? Just want to be sure we don't imply that users have to do something they don't.

README.md
tapashmajumder: Looks great. I would change 'deep linking enabled' to 'deferred deep linking' enabled.

bradumbaugh: Good catch—fixed in the suggestion!

README.md
bradumbaugh: Oh, I thought this all happened by default, but was customizable as necessary.

tapashmajumder: Yes we need to hook up silent push for ddl to work. I may have overlooked that earlier and caused some confusion.

PR #51, opened 2019-02-06
[MOB-179] - Persist triggers properly

tapashmajumder:


swift-sdk/Internal/InAppPersistence.swift
```diff
@@ -86,7 +152,7 @@ extension IterableInAppMessage : Codable {
         let messageId = (try? container.decode(String.self, forKey: .messageId)) ?? ""
         let campaignId = (try? container.decode(String.self, forKey: .campaignId)) ?? ""
         let channelName = (try? container.decode(String.self, forKey: .channelName)) ?? ""
-        let trigger = (try? container.decode(IterableInAppTriggerType.self, forKey: .trigger)) ?? .immediate
+        let trigger = (try? container.decode(IterableInAppTrigger.self, forKey: .trigger)) ?? .undefinedTrigger
```
vbabenkoru: When does it use `. undefinedTrigger`?

tapashmajumder: Here it will use undefined if for some reason the container is unable to deserialize. That should not happen in normal circumstances.

Normally use use the undefined when the string value sent by the server could not be parsed.

PR #55, opened 2019-02-13
[MOB-94] - Only set new email on updateEmail if it was using email.

tapashmajumder:


Tests/swift-sdk-swift-tests/IterableAPITests.swift
```diff
@@ -177,7 +177,7 @@ class IterableAPITests: XCTestCase {
                                     let body = networkSession.getRequestBody()
                                     TestUtils.validateElementPresent(withName: AnyHashable.ITBL_KEY_NEW_EMAIL, andValue: newEmail, inDictionary: body)
                                     TestUtils.validateElementPresent(withName: AnyHashable.ITBL_KEY_CURRENT_USER_ID, andValue: currentUserId, inDictionary: body)
-                                    XCTAssertEqual(IterableAPI.email, newEmail)
+                                    XCTAssertEqual(IterableAPI.userId, currentUserId)
```
vbabenkoru: Might also verify that email is nil, just in case. (not required for +1)

PR #57, opened 2019-02-19
[MOB-215] - Parse inAppType, contentType from payload

tapashmajumder:


Tests/swift-sdk-swift-tests/InAppHelperTests.swift
```diff
@@ -304,45 +304,151 @@ class InAppHelperTests: XCTestCase {
...
-            "channelName" : "channel2",
             "promoteToContent" : {
                 "title" : "title",
                 "subTitle" : "subtitle",
                 "imageUrl" : "http://somewhere.com/something.jpg"
             }
         }
         """
+        var customPayload2 = customPayloadStr2.toJsonDict()
+        customPayload2["inAppType"] = "inBox"
```
vbabenkoru: I think since the feature itself is `Inbox`, the internal name also should be `inbox` (all lowercase, without capitalization in the middle)

swift-sdk/IterableInApp.swift
```diff
@@ -42,10 +42,17 @@ open class DefaultInAppDelegate : IterableInAppDelegate {
     }
 }
 
+@objc
+public enum IterableInAppType : Int, Codable {
+    case `default`
+    case inBox
```
vbabenkoru: Same

PR #59, opened 2019-03-06
[MOB-233] - Support save to inbox

tapashmajumder:


vbabenkoru:
Let's merge this into a separate branch, like `inbox`, to make sure we can ship master when it is ready.

swift-sdk/IterableConfig.swift
```diff
@@ -35,16 +35,27 @@ import Foundation
...
+ * This protocol allows you to override default behavior when new in-app messages arrive.
  */
 @objc public protocol IterableInAppDelegate : class {
     /**
-     * This method is called when new inApp message is available.
+     * This method is called when inbox is ready. This is called right after app is launched and inbox messages
+     * are loaded from storage.
+     * - parameter messages: Array of `IterableInAppMessage` object containing information regarding inbox messages
+     */
+    @objc(onInboxReady:) func onInboxReady(messages: [IterableInAppMessage])
```
vbabenkoru: Rename to `onReady`? This also applies to in-app messages.

Tests/common/CommonExtensions.swift
```diff
@@ -48,7 +48,7 @@ extension IterableAPI {
                            networkSession: @escaping @autoclosure () -> NetworkSessionProtocol = MockNetworkSession(),
                            notificationStateProvider: NotificationStateProviderProtocol = SystemNotificationStateProvider(),
                            inAppSynchronizer: InAppSynchronizerProtocol = MockInAppSynchronizer(),
-                           inAppDisplayer: InAppDisplayerProtocol = MockInAppDisplayer(),
+                           iterableMessageDisplayer: IterableMessageDisplayerProtocol = MockIterableMessageDisplayer(),
```
vbabenkoru: I think `inAppDisplayer` is a better name if we're not planning to have generic messages for the time being.

swift-sdk/IterableConfig.swift
vbabenkoru: Actually we might not even need this if we have `onInboxChange`. It will run the first time we load the messages into memory.

swift-sdk/IterableConfig.swift
```diff
@@ -35,16 +35,27 @@ import Foundation
...
      * The default behavior is to `show` if you don't override this method.
      * - parameter message: `IterableInAppMessage` object containing information regarding inApp to display
      * - returns: Return `show` to show the inApp or `skip` to skip this.
      */
     @objc(onNewMessage:) func onNew(message: IterableInAppMessage) -> InAppShowResponse
+    /**
+     * This method is called when new inbox messages are available.
+     * - parameter inboxMessages: Array of `IterableInAppMessage` objects containing information regarding inbox messages
+     */
+    @objc(onNewInboxMessages:) func onNew(inboxMessages: [IterableInAppMessage])
```
vbabenkoru: `onInboxChanged`
We can add onAdded/onChanged/onRemoved later if users request it.

swift-sdk/IterableMessaging.swift
```diff
@@ -36,33 +56,35 @@ public protocol IterableInAppManagerProtocol {
...
-    case `default`
-    case inbox
+    public func onNew(inboxMessages: [IterableInAppMessage]) {
+        ITBInfo()
+    }
 }
 
 @objc
-public enum IterableInAppContentType : Int, Codable {
+public enum IterableContentType : Int, Codable {
```
vbabenkoru: If this can't apply to anything else, I think we can change it back to `IterableInAppContentType` to scope it to in-app.

swift-sdk/IterableMessaging.swift
```diff
@@ -36,33 +56,35 @@ public protocol IterableInAppManagerProtocol {
...
     case html
     case alert
     case banner
 }
 
 @objc
-public protocol IterableInAppContent {
-    var contentType: IterableInAppContentType {get}
+public protocol IterableContent {
```
vbabenkoru: Same as above, `IterableInAppContent`?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -297,16 +375,29 @@ extension InAppManager : InAppSynchronizerDelegate {
...
         ITBDebug()
         updateQueue.sync {
-            message.processed = processed
-            message.consumed = consumed
-            self.messagesMap.updateValue(message, forKey: message.messageId)
+            let toUpdate = message
+            if let didProcessTrigger = didProcessTrigger {
+                toUpdate.didProcessTrigger = didProcessTrigger
+            }
+            if let didProcessInbox = didProcessInbox {
```
vbabenkoru: Can we remove `didProcessInbox`? We don't stop processing them like in-app messages, and we probably know if there are changes when comparing the remote queue with the local queue, and can trigger updates based on that.

swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -1,105 +1,61 @@
...
-//  Created by Tapash Majumder on 6/7/18.
-//  Copyright © 2018 Iterable. All rights reserved.
+//  Created by Tapash Majumder on 3/11/19.
+//  Copyright © 2019 Iterable. All rights reserved.
 //
 
 import UIKit
 
-enum InAppNotificationType : Int {
+enum IterableMessageLocation : Int {
```
vbabenkoru: It's `Display type` in template UI.. I'm not sure, this one looks a bit generic, but it's up to you.

swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -1,105 +1,61 @@
...
-     Sets the padding
-     
-     - parameter: insetPadding the padding
-     
-     - remark: defaults to 0 for left/right if left+right > 100
-     */
-    func ITESetPadding(_ padding: UIEdgeInsets) {
-        var insetPadding = padding
+class IterableHtmlMessageViewController: UIViewController {
+    struct Input {
```
vbabenkoru: `settings`/`parameters`/etc?

swift-sdk/Internal/IterableMessageDisplayer.swift
```diff
@@ -0,0 +1,167 @@
...
+     - parameters:
+     - title:           The notification title
+     - body:            The notification message body
+     - buttonLeft:      The text of the left button
+     - buttonRight:     The text of the right button
+     - callbackBlock:   The callback to send after a button on the notification is clicked
+     
+     - remark:            passes the string of the button clicked to the callbackBlock
+     */
+    func showSystemNotification(_ title: String,
```
vbabenkoru: We might want to remove it.. not sure if anyone is using this.

swift-sdk/Internal/IterableUtil.swift
```diff
@@ -8,6 +8,12 @@
 import Foundation
 import os
 
+/// Functionality such as this will be built in for Swift 5.0. This will help with the transition
+enum IterableResult<T, E> {
```
vbabenkoru: 🎉

swift-sdk/Internal/IterableHtmlMessageViewController.swift
tapashmajumder: Lets deal with this one later.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -286,18 +302,21 @@ extension InAppManager : InAppSynchronizerDelegate {
...
             
             // add new ones
-            self.addNewMessages(messagesFromServer: messages)
+            let addedInboxCount = self.addNewMessages(messagesFromServer: messages)
             
             self.persister.persist(self.messagesMap.values)
             
-            // process inbox messages
-            self.processInboxMessages()
+            if deletedInboxCount + addedInboxCount > 0 {
```
vbabenkoru: Shouldn't we trigger `onInboxChanged` if the number of added and deleted messages is the same? (deleted one message, added one message)

swift-sdk/Internal/InAppManager.swift
```diff
@@ -445,19 +449,32 @@ extension InAppManager : InAppSynchronizerDelegate {
             return 0.0
         }
     }
-    
-    private func addNewMessages(messagesFromServer messages: [IterableInAppMessage]) {
+
+    // returns count of inbox messages (save to inbox)
+    private func addNewMessages(messagesFromServer messages: [IterableInAppMessage]) -> Int {
+        var count = 0
```
vbabenkoru: `inboxCount`?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -445,19 +449,32 @@ extension InAppManager : InAppSynchronizerDelegate {
...
             }
         }
+        return count
     }
     
-    private func removeDeletedMessages(messagesFromServer messages: [IterableInAppMessage]) {
-        getRemovedMessages(messagesFromServer: messages).forEach {
+    // return count of deleted inbox messages
+    private func removeDeletedMessages(messagesFromServer messages: [IterableInAppMessage]) -> Int {
+        var count = 0
```
vbabenkoru: `inboxCount`?

swift-sdk/Internal/InAppManager.swift
tapashmajumder: Not sure I understand. This will be triggered if any message was added or deleted. So if you delete one message and add another. deletedCount = 1, and addedCount = 1.

swift-sdk/IterableAPI.swift
```diff
@@ -45,6 +45,7 @@ public final class IterableAPI : NSObject {
                                   launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil,
                                   config: IterableConfig = IterableConfig()) {
         internalImplementation = IterableAPIInternal.initialize(apiKey: apiKey, launchOptions: launchOptions, config:config)
+        NotificationCenter.default.post(name: .iterableAppReady, object: self, userInfo: nil)
```
vbabenkoru: Why the change from `.synchronize` to `NotificationCenter`?

swift-sdk/Internal/InAppManager.swift
vbabenkoru: Oh, you're right. I thought the number of deleted entries was negative.

swift-sdk/Internal/InAppDisplayer.swift
```diff
@@ -51,12 +51,12 @@ class InAppDisplayer : InAppDisplayerProtocol {
             return false
         }
         
-        let input = IterableHtmlMessageViewController.Input(html: htmlString,
+        let parameters = IterableHtmlMessageViewController.Parameters(html: htmlString,
```
vbabenkoru: 👍

swift-sdk/IterableAPI.swift
tapashmajumder: Good question. Calling synchronize in initializer was causing some problems because `internalImplementation` was not set properly. This would show up only in unit tests where the app is being initialized more than once. But this is a better implementation than calling synchronize from inside the initializer.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -104,6 +149,26 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
             self.lastSyncTime = self.dateProvider.currentDate
         }
     }
+    
+    func set(read: Bool, forMessage message: IterableInAppMessage) {
```
vbabenkoru: Should we call `onInboxChanged` here, when read state changes?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -280,15 +361,16 @@ extension InAppManager : InAppSynchronizerDelegate {
         // We can only check applicationState from main queue so need to do the following
         DispatchQueue.main.async {
             if self.isOkToShowNow(message: message) {
-                self.updateMessage(message, processed: true)
+                self.updateMessage(message, didProcessTrigger: true)
```
vbabenkoru: 👍

swift-sdk/IterableMessaging.swift
```diff
@@ -121,47 +156,57 @@ public final class IterableInAppMessage : NSObject {
...
-    public let inAppType: IterableInAppType
-    
     /// when to trigger this in-app
     public let trigger: IterableInAppTrigger
     
     /// when to expire this in-app, nil means do not expire
     public let expiresAt: Date?
     
-    /// The content of the inApp message
+    /// The content of the inbox message
```
vbabenkoru: Not necessarily inbox, `IterableInAppContent` is common for any in-app messages.

swift-sdk/IterableMessaging.swift
```diff
@@ -121,47 +156,57 @@ public final class IterableInAppMessage : NSObject {
...
 
-    /// Whether we have processed this message.
+    /// Whether we have processed the trigger for this message.
+    /// Note: This is internal and not public
+    internal var didProcessTrigger = false
+    
+    /// Whether we have processed 'save to inbox' for this message.
     /// Note: This is internal and not public
-    var processed: Bool = false
+    internal var didProcessInbox = false
```
vbabenkoru: It doesn't look like `didProcessInbox` is used anywhere anymore.. can we remove it?

swift-sdk/Internal/InAppManager.swift
tapashmajumder: We are using onInboxChanged for passing changes coming from the server to the client. I don't think this will be helpful because the client that will set the value to read will get a onInboxChanged callback again. 
I can see the value if different parts of the client iOS app is using this callback. But for that a notification is more appropriate.

swift-sdk/IterableMessaging.swift
tapashmajumder: oops. Yes need to remove this.

swift-sdk/IterableAPI.swift
vbabenkoru: `.synchronize` seemed a bit more explicit to me, even though it is a direct call. A `NotificationCenter` message is something that's global and can even be sent by anyone.

swift-sdk/Internal/InAppManager.swift
tapashmajumder: Now we have notification. I have added the call to fire this notification.

tapashmajumder:
I created a separate issue for this. Hopefully I will look into this soon.
https://iterable.atlassian.net/browse/MOB-279

PR #61, opened 2019-03-07
[MOB-241] - InboxManager related changes

tapashmajumder:


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -710,6 +713,7 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
                                         dateProvider: dateProvider,
                                         retryInterval: config.inAppDisplayInterval)
         self.inAppManager = inAppManager
+        self.inboxManager = inAppManager
```
davidtruong: Should this be inboxManager?

swift-sdk/IterableConfig.swift
```diff
@@ -35,18 +35,35 @@ import Foundation
...
+     * This method is called when new in-app message is available.
      * The default behavior is to `show` if you don't override this method.
      * - parameter message: `IterableInAppMessage` object containing information regarding inApp to display
      * - returns: Return `show` to show the inApp or `skip` to skip this.
      */
     @objc(onNewMessage:) func onNew(message: IterableInAppMessage) -> InAppShowResponse
 }
 
+/**
+ * This protocol informs you when when new inbox messages arrive.
```
davidtruong: extra when in comment

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: inAppManager to the right is internal class which implements both inboxManager protocol and inAppManager protocol. In LHS we have external inBoxManager and inAppManager.

So inAppManager in LHS is different than inAppManager in RHS.

PR #64, opened 2019-03-27
[MOB-280] - Delete and dismiss custom action.

tapashmajumder:


swift-sdk/ITBConsts.swift
```diff
@@ -168,3 +168,8 @@ public enum JsonKey : String {
     case inboxSubtitle = "subTitle"
     case inboxIcon = "icon"
 }
+
+public enum IterableInAppCustomActionName : String, CaseIterable {
+    case dismiss = "itbl_inapp_dismiss"
```
vbabenkoru: `itbl://itbl_inapp_dismiss` seems too long.. can we use just `itbl://dismiss` and `itbl://delete`? Until we have an editing UI where you can select these things from a combo box, users will have to type this in the URL field, and a shorter and clearer name would be easier for a marketer to use. That would prevent the app from using these names for their own actions, but as long as we document those reserved names, I think it should be fine..

tapashmajumder: I am not sure about that. We should have `itbl_` namespace for the custom action at least. If we don't then they can define a custom action called 'delete' and there will be  namespace collision.

vbabenkoru: As an alternative, we can switch to using `action://` or `custom://` for app custom actions, and use `itbl://` only for SDK-defined actions.

tapashmajumder: Yes I like that. One thing we should keep in mind is it appears we always prepend itbl:// when there is no url specified. We should probably prepend action://

vbabenkoru: If the custom action is empty, nothing will happen either way, right?

davidtruong: I agree that we should use `itbl:// ` for sdk-defined actions.
That way we can support things such as going to the next step of a multi-step in-app rather than automatically closing the dialog after the url click.

tapashmajumder:
I am closing/deleting this. There is another branch and PR for this.

PR #72, opened 2019-04-02
[MOB-280] - itbl://dismiss, delete etc.

tapashmajumder:


swift-sdk/Internal/InAppManager.swift
```diff
@@ -241,9 +241,18 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
...
+    private func handleIterableCustomAction(name: String, forMessage message: IterableInAppMessage) {
+        guard let iterableCustomActionName = IterableCustomActionName(rawValue: name) else {
+            return
+        }
+
+        switch iterableCustomActionName {
+        case .delete:
+            remove(message: message)
+            break
+        case .dismiss:
```
davidtruong: @tapashmajumder Since the message is now persisted for the inbox, does this also handle the case so that we won't display the message again as in in-app?

tapashmajumder: @davidtruong Yes, the message should be marked as processed.

PR #80, opened 2019-04-22
[MOB-279] - Fix init

tapashmajumder:


swift-sdk/Internal/InAppManager.swift
```diff
@@ -515,9 +383,145 @@ extension InAppManager : InAppSynchronizerDelegate {
...
+    private let scheduleQueue = DispatchQueue(label: "ScheduleQueue")
+    private let syncQueue = DispatchQueue(label: "SyncQueue")
+    private let callbackQueue = DispatchQueue(label: "CallbackQueue")
+    private var lastSyncTime: Date? = nil
+    private let moveToForegroundSyncInterval: Double = 1.0 * 60.0 // don't sync within sixty seconds
+}
+
+extension InAppManager : InAppNotifiable {
+    func onInAppSyncNeeded() {
+        ITBInfo()
```
davidtruong: This will log twice since we already call ITBInfo in synchronize.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -515,9 +383,145 @@ extension InAppManager : InAppSynchronizerDelegate {
             }
         }
     }
+
+    private func handle(clickedUrl url: URL?, forMessage message: IterableInAppMessage) {
+        guard let theUrl = url, let inAppClickedUrl = InAppHelper.parse(inAppUrl: theUrl) else {
```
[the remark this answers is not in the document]
tapashmajumder: These are two separate variables even if we name the same. I can use the same variable name `url` just to be consistent but it is a different variable than the original `url`.
I will change the var name.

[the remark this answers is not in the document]
tapashmajumder: Oh I see the problem with using the same name is that we now lose the original optional url for error message in the next line.

swift-sdk/Internal/InAppManager.swift
tapashmajumder: We can call synchronize from other places. This log will help us trace whether we are calling synchronize from this method or some other method.

swift-sdk/Internal/InAppInternal.swift
```diff
@@ -5,59 +5,42 @@
...
     }
-    
+
     deinit {
         ITBInfo()
     }
     
     // how many messages to fetch
-    private let numMessages = 10
+    private let numMessages = 100
```
vbabenkoru: Do we want to change this in the current beta (6.1.0)?

swift-sdk/Internal/InAppInternal.swift
```diff
@@ -5,59 +5,42 @@
...
+class InAppSynchronizer : InAppSynchronizerProtocol {
     init() {
         ITBInfo()
     }
     
-    func sync() {
+    func sync() -> Future<[IterableInAppMessage]> {
         ITBInfo()
-        guard let internalApi = self.internalApi else {
+        guard let internalApi = IterableAPI.internalImplementation else {
```
vbabenkoru: Hmm, this removes dependency injection in favor of accessing a global property..

swift-sdk/Internal/InAppInternal.swift
```diff
@@ -5,59 +5,42 @@
...
 extension IterableInAppTriggerType {
     static let defaultTriggerType = IterableInAppTriggerType.immediate // default is what is chosen by default
     static let undefinedTriggerType = IterableInAppTriggerType.never // undefined is what we select if payload has new trigger type
 }
 
-class InAppSilentPushSynchronizer : InAppSynchronizerProtocol {
-    weak var internalApi: IterableAPIInternal?
-    weak var inAppSyncDelegate: InAppSynchronizerDelegate?
-    
+class InAppSynchronizer : InAppSynchronizerProtocol {
```
vbabenkoru: Looking at the new implementation, this class isn't doing any synchronization and just returns a server-side queue now. What do you think about renaming it to something like `InAppFetcher`?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -182,6 +162,39 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
         }
     }
     
+    private func synchronize() {
+        ITBInfo()
+        syncQueue.async {
+            self.synchronizer.sync()
```
vbabenkoru: Yeah, since this doesn't actually sync anything, I think it should be renamed.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -517,9 +385,145 @@ extension InAppManager : InAppSynchronizerDelegate {
...
+    func onInAppSyncNeeded() {
+        ITBInfo()
+        synchronize()
+    }
+    
+    // from server side
+    func onInAppRemoved(messageId: String) {
+        ITBInfo()
+        
+        updateQueue.async {
```
vbabenkoru: Should this post a `iterableInboxChanged` notification?

swift-sdk/IterableAPI.swift
```diff
@@ -44,8 +44,8 @@ public final class IterableAPI : NSObject {
     public static func initialize(apiKey: String,
                                   launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil,
                                   config: IterableConfig = IterableConfig()) {
-        internalImplementation = IterableAPIInternal.initialize(apiKey: apiKey, launchOptions: launchOptions, config:config)
-        NotificationCenter.default.post(name: .iterableAppReady, object: self, userInfo: nil)
+        internalImplementation = IterableAPIInternal(apiKey: apiKey, launchOptions: launchOptions, config:config)
+        internalImplementation?.start()
```
vbabenkoru: 👍

swift-sdk/IterableAPI.swift
```diff
@@ -498,6 +498,7 @@ public final class IterableAPI : NSObject {
     @objc public static var inAppManager: IterableInAppManagerProtocol {
         guard let internalImplementation = internalImplementation else {
             ITBError("IterableAPI is not initialized yet. InApp will not work now.")
+            assertionFailure("IterableAPI is not initialized yet. InApp will not work now.")
```
vbabenkoru: `In-app`
Will this crash the app?

swift-sdk/IterableInboxViewController.swift
```diff
@@ -144,6 +144,7 @@ open class IterableInboxViewController: UITableViewController {
                 cell.iconImageView?.image = UIImage(data: data)
             } else {
                 cell.iconImageView?.backgroundColor = UIColor(hex: "EEEEEE") // loading image
+                cell.iconImageView?.image = nil
```
vbabenkoru: Looks like an unrelated change.. Is it intentional?

swift-sdk/Internal/InAppInternal.swift
tapashmajumder: Yes we probably should. 
Later on, if needed, maybe we should think about having a config parameter.

vbabenkoru: Yeah, 10 messages is low enough that some customers might hit that limit during implementation..

swift-sdk/Internal/InAppInternal.swift
tapashmajumder: I will revert back to field injection instead of global.

swift-sdk/IterableAPI.swift
tapashmajumder: Since this is assertionFailure() this will only crash in debug builds. It will be replaced with a noop in production builds.

swift-sdk/IterableInboxViewController.swift
tapashmajumder: Yes. I think you already made this change in another branch.

swift-sdk/Internal/InAppManager.swift
tapashmajumder: Awesome catch!

swift-sdk/IterableAPI.swift
vbabenkoru: `InApp` -> `In-app`

PR #81, opened 2019-04-26
[MOB-351] - Update tableview using diff util

tapashmajumder:


swift-sdk/IterableInboxViewController.swift
```diff
@@ -89,13 +93,15 @@ open class IterableInboxViewController: UITableViewController {
 
     private let iterableCellNibName = "IterableInboxCell"
     
-    private var viewModels = [InboxMessageViewModel]()
-
+    private lazy var diffCalculator: TableViewDiffCalculator<Int, InboxMessageViewModel>? = {
+        TableViewDiffCalculator(tableView: self.tableView, initialSectionedValues: SectionedValues([]))
```
vbabenkoru: If we only have one section, it might be easier/cleaner to use `SingleSectionTableViewDiffCalculator` instead.

swift-sdk.xcodeproj/project.pbxproj
```diff
@@ -585,6 +594,17 @@
 			path = "swift-sdk-objc-tests";
 			sourceTree = "<group>";
 		};
+		AC819182227138D40014955E /* Diff */ = {
```
vbabenkoru: I think we should move these library files from Dwifft to a separate folder in the filesystem as well.

PR #86, opened 2019-05-02
[MOB-339] - Nav Controller

tapashmajumder:
Inbox Nav controller.

### DIFF
--- a/swift-sdk/IterableInboxNavigationViewController.swift
+++ b/swift-sdk/IterableInboxNavigationViewController.swift
@@ -17,12 +17,13 @@ open class IterableInboxNavigationViewController: UINavigationController {
             }
         }
     }
-
+    
     // MARK: Initializers
     
     /// This initializer should be used when initializing from Code.
     public init() {
-        super.init(rootViewController: IterableInboxViewController(style: .plain))
+        ITBInfo()
+        super.init(nibName: nil, bundle: nil)
         setup()
     }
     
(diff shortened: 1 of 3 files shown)


swift-sdk/IterableInboxNavigationViewController.swift
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+    }
+
+    // MARK: Initializers
+    
+    /// This initializer should be used when initializing from Code.
+    public init() {
+        super.init(rootViewController: IterableInboxViewController(style: .plain))
+        setup()
```
vbabenkoru: Since we're initializing with an existing instance of an `IterableInboxViewController`, won't `setup()` log an error because there is already a view controller in the stack? (line 59)

swift-sdk/IterableInboxViewController.swift
```diff
@@ -40,6 +40,9 @@ open class IterableInboxViewController: UITableViewController {
         tableView.rowHeight = UITableView.automaticDimension
         tableView.estimatedRowHeight = UITableView.automaticDimension
         
+        // Set footer view so that
```
vbabenkoru: Unfinished comment :)

swift-sdk/IterableInboxNavigationViewController.swift
tapashmajumder: Good catch!

PR #93, opened 2019-05-08
[MOB-380] iterable namespace fix for master

tapashmajumder:
use itbl:// for backward compatibility.


swift-sdk/Internal/InAppHelper.swift
```diff
@@ -472,8 +474,9 @@ struct InAppHelper {
     
     private enum UrlScheme : String {
         case applewebdata = "applewebdata"
-        case itbl = "itbl"
+        case iterable = "iterable"
         case action = "action"
+        case backwardCompat = "itbl"
```
vbabenkoru: I think it might be a bit clearer to name this itbl and add a comment that it's for backwards compatibility since all the other values are following the convention of using the scheme name itself.

tapashmajumder: Will fix this.

PR #101, opened 2019-05-21
Push auth request

bradumbaugh:
Added information about requesting permission to receive push notifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ IterableAppExtensions.framework
 
     - By default, the SDK handles deep links with the the URL delegate
     assigned to `IterableConfig`. Follow the instructions in 
-    [Deep Linking](#deep-linking) to migrate any existing URL handling code 
+    [Deep Links](#deep-links) to migrate any existing URL handling code 
     to this new API.
 
 ## Sample projects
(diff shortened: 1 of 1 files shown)


bradumbaugh:
@tapashmajumder Not sure about this build failure?

README.md
```diff
@@ -186,9 +186,47 @@ IterableAPI.email = @"user@example.com";
...
+    [center requestAuthorizationWithOptions: (UNAuthorizationOptionAlert + UNAuthorizationOptionBadge + UNAuthorizationOptionSound)
+       completionHandler:^(BOOL granted, NSError * _Nullable error) {
+       // ...
+    }];
+    ```
+    
+    By default, if you have set `IterableAPI.email` or `IterableAPI.userId`
+    (as described above), the SDK will then register the app for remote 
+    notifications automatically.
+
```
tapashmajumder: This is a little tricky. If you set userId or email. IterableSDK will register for token without asking for permission. If the app has background mode enabled then we can send silent pushes. I am not sure if we should make this clearer.

tapashmajumder: This is a constant source of confusion.

bradumbaugh: @tapashmajumder I updated this—could you take another look?

bradumbaugh:
@tapashmajumder 

- I added some content about `autoPushRegistration`. Let me know if this looks correct to you.
- Doing so pushed me to move more content from the "using the SDK" section into the "configuring the SDK" section. I'm trying to make the "configuring the SDK" section all about setup, and the "using the SDK" all about displaying messages, handling user interactions, etc.
- Other minor fixes throughout.
- Again, this is a fairly significant update. Apologies. Happy to talk in person about it if that's easier.

tapashmajumder:
+1

PR #103, opened 2019-05-22
[MOB-432] - refactor inapp displayer

tapashmajumder:


tapashmajumder:
This is pretty much same as before. Previously a value of false/true returned right away and then callback called on user click.
Now we return an error right away (== false earlier) then a Future is returned instead of callback. We can look maybe making this better in the future but this is no worse that what is before.

Returning a Future instead of callbacks will help us refactor the code better down the line. This relates to another dependent branch that I am working on.

vbabenkoru:
The return value and the callback had different purpose and context though...

tapashmajumder:
What do you think about returning ShowResult? This is a `sum type` depending on the outcome. I think that is a decent approach.

enum ShowResult {
  case opened(Future<URL, Error>) // if opened returned future
  case notOpened // if not opened nothing else is needed
}

vbabenkoru:
Yeah, I think that's a good idea.

swift-sdk/Internal/InAppDisplayer.swift
```diff
@@ -31,41 +35,39 @@ class InAppDisplayer : InAppDisplayerProtocol {
...
     @discardableResult static func showIterableHtmlMessage(_ htmlString: String,
                                                            trackParams: IterableNotificationMetadata? = nil,
                                                            backgroundAlpha: Double = 0,
-                                                           padding: UIEdgeInsets = .zero,
-                                                           callbackBlock: ITBURLCallback?
-        ) -> Bool {
+                                                           padding: UIEdgeInsets = .zero) -> ShowResult {
         guard let topViewController = getTopViewController() else {
-            return false
+            return .notShown("No top ViewController.")
```
vbabenkoru: This is great

swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -17,18 +17,18 @@ class IterableHtmlMessageViewController: UIViewController {
     struct Parameters {
         let html: String
         let padding: UIEdgeInsets
-        let callback: ITBURLCallback?
+        let result: Promise<URL, IterableError>
```
vbabenkoru: Invert the control flow here: instead of taking `result` as a parameter, return it as a result of `show`, etc.

swift-sdk/Internal/InAppDisplayer.swift
```diff
@@ -31,41 +35,39 @@ class InAppDisplayer : InAppDisplayerProtocol {
...
         }
         if topViewController is IterableHtmlMessageViewController {
-            ITBError("Skipping the in-app notification. Another notification is already being displayed.")
-            return false
+            return .notShown("Skipping the in-app notification. Another notification is already being displayed.")
         }
         
+        let result = Promise<URL, IterableError>()
```
vbabenkoru: Same

PR #109, opened 2019-06-03
[MOB-442] - refactor inappmanager

tapashmajumder:


swift-sdk/Internal/InAppManager.swift
```diff
@@ -533,18 +488,21 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
...
+    private let syncQueue = DispatchQueue(label: "SyncQueue")
+    
+    private var syncResult: Future<Bool, Error>?
     private var lastSyncTime: Date? = nil
     private let moveToForegroundSyncInterval: Double = 1.0 * 60.0 // don't sync within sixty seconds
 }
 
 extension InAppManager : InAppNotifiable {
-    func onInAppSyncNeeded() {
+    func onInAppSyncNeeded() -> Future<Bool, Error> {
```
vbabenkoru: Since the purpose of the method has changed and it returns a Future to the caller, `onInAppSyncNeeded` isn't a proper name anymore I think.. It should be a verb like `scheduleSync`, etc.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -136,62 +133,140 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
...
+        switch processMessagesResult {
+        case .noShow(messagesMap: let messagesMap):
+            return messagesMap
+        case .show(message: _, messagesMap: let messagesMap):
+            return messagesMap
+        }
+    }
+    
+    // Not a pure function.
+    private func handleShowMessage(fromProcessMessagesResult processMessagesResult: ProcessMessagesResult) {
```
vbabenkoru: Why `handleShowMessage` instead of just `showMessage`?

swift-sdk/Internal/InAppManager.swift
```diff
@@ -136,62 +133,140 @@ class InAppManager : NSObject, IterableInAppManagerProtocolInternal {
...
+                if syncResult.isResolved() {
+                    self.syncResult = self.synchronize()
+                } else {
+                    self.syncResult = syncResult.flatMap { _ in self.synchronize() }
+                }
+            } else {
+                self.syncResult = self.synchronize()
+            }
+            self.syncResult?.onSuccess { _ in
+                result.resolve(with: true)
```
vbabenkoru: This appears to only resolve the Future with a successful result, but not pass errors.

swift-sdk/Internal/EmptyInAppManager.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    
+    func set(read: Bool, forMessage message: IterableInAppMessage) {
+    }
+    
+    func getUnreadInboxMessagesCount() -> Int {
+        return 0
+    }
+    
+    func onInAppSyncNeeded() -> Future<Bool, Error> {
```
vbabenkoru: What is this `Future` used for?

tapashmajumder: Signifies completion of sync. It will be mostly ignored.

vbabenkoru: Is it only ever used in tests? I'm not a big fan of complicating code to simplify tests, but it's up to you.

tapashmajumder: This is needed for more than tests. If a sync request arrives before another finishes, this chains the requests.

PR #110, opened 2019-06-11
[MOB-340] Tests for Inbox

roninopf:


Tests/swift-sdk-swift-tests/IterableInboxViewControllerTests.swift
```diff
@@ -0,0 +1,35 @@
...
+            return
+        }
+        
+        XCTAssertEqual(inboxViewController4.tableView.numberOfSections, 1)
+    }
+    
+//    private func createDefaultContent() -> IterableInAppContent {
+//        return IterableHtmlInAppContent(edgeInsets: .zero, backgroundAlpha: 0.0, html: "")
+//    }
+}
```
tapashmajumder: This seems to be left over from somewhere else.

Tests/ui-tests/IterableInboxViewControllerUITests.swift
```diff
@@ -0,0 +1,48 @@
...
+    override func setUp() {
+        continueAfterFailure = false
+    }
+    
+    func testRowSelect() {
+        app.buttons["Show Inbox"].tap()
+        
+        sleep(1)
+        
+//        let inboxTableView = app.tables.element(boundBy: 0)
```
tapashmajumder: Please remove unused comments.
Also I am not sure what this test is doing.

Tests/ui-tests/IterableInboxViewControllerUITests.swift
```diff
@@ -0,0 +1,48 @@
...
+        
+        
+        
+//        XCTAssertFalse()
+    }
+    
+    
+    
+    private func createDefaultContent() -> IterableInAppContent {
```
tapashmajumder: What is this method used for?

Tests/swift-sdk-swift-tests/IterableInboxNavigationViewControllerTests.swift
```diff
@@ -0,0 +1,38 @@
...
+        XCTAssertNotNil(inboxNavigationVC.viewControllers[0] as? IterableInboxViewController)
+    }
+    
+    func testInitWithCoder() {
+        guard let inboxNavigationVC = IterableInboxNavigationViewController(coder: NSKeyedUnarchiver(forReadingWith: Data())) else {
+            XCTFail()
+            return
+        }
+        
+        XCTAssertNil(inboxNavigationVC.navigationController?.tabBarItem.badgeValue)
```
vbabenkoru: This will always be nil, since `inboxNavigationVC` has no parent navigation controller.

Tests/swift-sdk-swift-tests/IterableInboxViewControllerTests.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+
+import XCTest
+
+@testable import IterableSDK
+
+class IterableInboxViewControllerTests: XCTestCase {
+    func testInitializers() {
+        let inboxViewController1 = IterableInboxViewController()
+        XCTAssertEqual(inboxViewController1.tableView.numberOfSections, 1)
```
vbabenkoru: Not sure if we need to check the number of sections for all these cases.

Tests/ui-tests/IterableInboxViewControllerUITests.swift
roninopf: Originally it was used to test making new in app messages so it was a shortcut method to fill them out.

Tests/swift-sdk-swift-tests/IterableInboxViewControllerTests.swift
roninopf: Yeah, I wasn't entirely sure the way to do this, but my objective for those checks was to ensure it actually created the full view controller. It's probably not necessary though, you're right. I can just check if the VC isn't nil.

Tests/swift-sdk-swift-tests/ClassExtensionsTests.swift
```diff
@@ -0,0 +1,24 @@
...
+
+import XCTest
+
+@testable import IterableSDK
+
+class ClassExtensionsTests: XCTestCase {
+    func testUIColorInit() {
+        let blackColor = UIColor(red: 0, green: 0, blue: 0, alpha: 1)
+        
+        guard let hexColor = UIColor(hex: "000000") else {
```
davidtruong: For clarity this should either be of the format "#000000" or 0x000000.

roninopf: I like that idea but the current initializer doesn't support it, but it could be modified in the future to accommodate those common forms of representation

PR #112, opened 2019-06-13
[MOB-471] - We need to show read/unread properly when cell is highlighted.

tapashmajumder:


swift-sdk/IterableInboxCell.swift
```diff
@@ -38,4 +38,15 @@ open class IterableInboxCell: UITableViewCell {
         // Initialization code
     }
 
+    // override this to show unreadCircle color when highlighted
+    // otherwise the background color is not correct.
+    open override func setHighlighted(_ highlighted: Bool, animated: Bool) {
+        let color = unreadCircleView?.backgroundColor
```
davidtruong: Can you move this def of color inside of the if highlighted block?

tapashmajumder: We can't add this to the `if highlighted` block because otherwise the original color will be overwritten. We have to save the original color before calling the super method.

davidtruong: Got it

PR #116, opened 2019-06-19
Change header to use Api-Key instead of api_key for POSTs

roninopf:
Tests were also adjusted to reflect the use of the new key.

### DIFF
--- a/swift-sdk/ITBConsts.swift
+++ b/swift-sdk/ITBConsts.swift
@@ -34,6 +34,7 @@ public extension String {
 
 // Keys
 public extension AnyHashable {
+    static let ITBL_HEADER_API_KEY = "Api-Key"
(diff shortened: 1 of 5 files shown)


swift-sdk/ITBConsts.swift
```diff
@@ -34,6 +34,7 @@ public extension String {
 
 // Keys
 public extension AnyHashable {
+    static let ITBL_HEADER_API_KEY = "Api-Key"
```
davidtruong: What's the effect of this change? What behavior do old sdks experience when using the old header?

roninopf: It's for RFC compliance, so there's no technical difference between the before and after of this change (since the backend has supported both for a while).

PR #119, opened 2019-06-25
Updates for Carthage/CocoaPods instructions

bradumbaugh:
- Created a section for "Installing with a dependency manager", and put both the CocoaPods and the Carthage sections there
- Clarified the Carthage and the CocoaPods instructions a bit (reorganized some steps, updated text, etc)


bradumbaugh:
@roninopf When reviewing this PR, please in particular let me know if step #8 for Carthage is correct (as we discussed internally).

PR #124, opened 2019-07-17
[MOB-505] add fields to open and click requests

roninopf:
There's also a bunch of spacing and some casing standardization I did with associated files; definitely set "hide whitespace changes" to true.

Top level changes:
- split `IterableNotificationMetadata` into `IterablePushNotificationMetadata` and `IterableInAppMessageMetadata` and adjusted code using those respectively
- added `saveToInbox`, `trigger`, `location` fields to `trackInAppOpen` and `trackInAppClick` calls, as optionals and defaulting fields through all the levels the API call gets made

Contributing changes:
- `deviceId` gets passed through
- new static constants made


vbabenkoru:
@roninopf Could you extract the refactor/formatting/cleanup part into a separate PR?

roninopf:
> @roninopf Could you extract the refactor/formatting/cleanup part into a separate PR?

Yep, finally got it separated out into this PR: https://github.com/Iterable/swift-sdk/pull/125

And this PR is based off of that PR.

swift-sdk/Internal/ApiClient.swift
```diff
@@ -48,12 +48,13 @@ protocol AuthProvider: class {
...
 
-class ApiClient : ApiClientProtocol {
-    init(apiKey: String, authProvider: AuthProvider, endPoint: String, networkSession: NetworkSessionProtocol) {
+class ApiClient: ApiClientProtocol {
+    init(apiKey: String, authProvider: AuthProvider, endPoint: String, networkSession: NetworkSessionProtocol, deviceId: String) {
         self.apiKey = apiKey
         self.authProvider = authProvider
         self.endPoint = endPoint
         self.networkSession = networkSession
+        self.deviceId = deviceId
```
vbabenkoru: `deviceId` seems too specific to add here.. i.e. even the auth part is abstracted into a provider. What do you think about either passing deviceId explicitly (like it is now in `createRegisterTokenRequest`), or creating another provider?
`deviceId` is also used in `createRegisterTokenRequest` so whatever we do with `deviceId`, it should be the same in `createRegisterTokenRequest` and `track*`.

swift-sdk/IterableNotificationMetadata.swift
```diff
@@ -8,68 +8,59 @@
 
 import Foundation
 
-/**
- `IterableNotificationMetadata` represents the metadata in an Iterable push notification
- */
-@objc public class IterableNotificationMetadata: NSObject {
```
vbabenkoru: This class is part of public API, this change will remove `isRealCampaignNotification` from `IterablePushNotificationMetadata`.

swift-sdk/IterableNotificationMetadata.swift
```diff
@@ -8,68 +8,59 @@
...
-    @objc public var campaignId: NSNumber = NSNumber(value: 0)
-    @objc public var templateId: NSNumber? = nil
-    @objc public var messageId: String? = nil
+@objc public class IterableInAppMessageMetadata: IterableNotificationMetadata {
```
vbabenkoru: Does this have to be a subclass of `IterableNotificationMetadata`? Looks like it's not setting the parent fields in the constructor.

swift-sdk/IterableNotificationMetadata.swift
vbabenkoru: Disregard this, this class should actually be private as it's not returned by any of the public methods.

swift-sdk/Internal/ApiClient.swift
```diff
@@ -47,12 +47,12 @@ protocol AuthProvider: class {
...
 
 class ApiClient: ApiClientProtocol {
-    init(apiKey: String, authProvider: AuthProvider, endPoint: String, networkSession: NetworkSessionProtocol, deviceId: String) {
+    init(apiKey: String, authProvider: AuthProvider, endPoint: String, networkSession: NetworkSessionProtocol, deviceMetadata: DeviceMetadata) {
         self.apiKey = apiKey
         self.authProvider = authProvider
         self.endPoint = endPoint
         self.networkSession = networkSession
-        self.deviceId = deviceId
+        self.deviceMetadata = deviceMetadata
```
vbabenkoru: Hmm I think it's better to move deviceMetadata to request parameters instead of `ApiClient` since device metadata does not apply to all or most of the API calls.

swift-sdk/Internal/ApiClient.swift
```diff
@@ -30,7 +30,7 @@ protocol ApiClientProtocol: class {
     
     func track(inAppOpen messageId: String, saveToInbox: Bool?, silentInbox: Bool?, location: String?) -> Future<SendRequestValue, SendRequestError>
     
-    func track(inAppClick messageId: String, saveToInbox: Bool?, silentInbox: Bool?, location: String?, buttonURL: String) -> Future<SendRequestValue, SendRequestError>
+    func track(inAppClick messageId: String, saveToInbox: Bool?, silentInbox: Bool?, location: String?, clickedUrl: String) -> Future<SendRequestValue, SendRequestError>
```
vbabenkoru: Move `deviceMetadata` here

swift-sdk/Internal/RequestCreator.swift
```diff
@@ -311,22 +311,19 @@ struct RequestCreator {
             context[.ITBL_IN_APP_LOCATION] = location
         }
         
-        context[String.ITBL_DEVICE_INFO_KEY] = getDeviceInfo(deviceId)
+        context[String.ITBL_DEVICE_INFO_KEY] = translateDeviceMetadata(metadata: deviceMetadata)
         
         dict["messageContext"] = context
```
vbabenkoru: Extract into constants

PR #130, opened 2019-08-07
Merge header fix

roninopf:


swift-sdk/Internal/IterableRequestUtil.swift
```diff
@@ -9,23 +9,24 @@
...
     
-    static func createPostRequest(forApiEndPoint apiEndPoint: String, path: String, args: [String: String]? = nil, body: Data? = nil) -> URLRequest? {
-        guard let url = getUrlComponents(forApiEndPoint: apiEndPoint, path:path, args: args)?.url else {
+    static func createPostRequest(forApiEndPoint apiEndPoint: String, path: String, apiKey: String, args: [String: String]? = nil, body: Data? = nil) -> URLRequest? {
+        guard let url = getUrlComponents(forApiEndPoint: apiEndPoint, path: path, args: args)?.url else {
             return nil
         }
         
         var request = URLRequest(url: url)
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
```
tapashmajumder: IterableRequestUtil is a general class. So the responsibility of this should just be to create a URL request given the parameters. The header fields should be passed to it instead of hardcoding the header parameters here.

So this method should take header values as parameters.

The responsibility to create iterable specific request items (headers etc) should go in RequestCreator.swift file. This should create a GET or POST request with headers.

tapashmajumder: One other way could be to merge this pull request as is. Then I can create a separate PR to refactor what I mentioned above. I think I like that approach better.

roninopf: Yeah, let's do that refactor in another PR; this one is specifically for just merging back changes from master anyways.

tapashmajumder:
+1

PR #131, opened 2019-08-07
[mob-598] - format using swift formatter

tapashmajumder:


roninopf:
A LOT of it looks great, but I'm interested in knowing if we can fix:
- the indentation removal on blank lines; I like having the indentation so that adding in new code is easier by not needing to hit tab to get to the proper level
- putting certain files on an avoid list; namely the ones related to `Dwifft`

PR #133, opened 2019-08-09
[MOB-607] - Refactored code to add `inboxMode`.

tapashmajumder:
Currently only inboxMode = .nav is supported.


tapashmajumder:
This PR is really simple just adds the hooks to handle different modes for inbox.

PR #146, opened 2019-08-27
[MOB-647] test getInAppMessages request

roninopf:


Tests/swift-sdk-swift-tests/RequestCreatorTests.swift
```diff
@@ -29,6 +29,59 @@ class RequestCreatorTests: XCTestCase {
...
+        guard let header = urlRequest.allHTTPHeaderFields else {
+            XCTFail("no header")
+            return
+        }
+        
+        XCTAssertEqual(header[AnyHashable.ITBL_HEADER_SDK_PLATFORM], String.ITBL_PLATFORM_IOS)
+        XCTAssertEqual(header[AnyHashable.ITBL_HEADER_SDK_VERSION], IterableAPI.sdkVersion)
+        XCTAssertEqual(header[AnyHashable.ITBL_HEADER_API_KEY], apiKey)
+        
+        do {
```
tapashmajumder: Not needed for +1 but consider replacing do..catch with guard and try?
Usually try/catch semantics are not that great for Swift. That is why we have a soft try? which returns nil when it fails.

tapashmajumder: so the code would look like
`guard let success = try? request else {
XCTFail(...) 
return
}
`

Tests/swift-sdk-swift-tests/RequestCreatorTests.swift
```diff
@@ -29,6 +29,50 @@ class RequestCreatorTests: XCTestCase {
...
+        
+        if let _ = try? failingRequest.get() {
+            XCTFail("request succeeded despite userId and email being nil")
+        }
+    }
+    
+    func testGetInAppMessagesRequest() {
+        let inAppMessageRequestCount: NSNumber = 42
+        
+        let request = createRequestCreator().createGetInAppMessagesRequest(inAppMessageRequestCount)
```
tapashmajumder: `guard let request = createRequestCreator().createGetInAppMessagesRequest(inAppMessageRequestCount) else {
XCTFail("Could not create getInAppMessagesRequest")
return
}`

tapashmajumder: Suggestion:
`        guard case let .success(request) = createRequestCreator().createGetInAppMessagesRequest(2) else {
            XCTFail("Could not create getInAppMessagesRequest")
            return
        }
        guard case let .get(getRequest) = request, let args = getRequest.args else {
            XCTFail("could not create getRequest")
            return
        }
`

tapashmajumder: Now you don't need line numbers 61-69. No need for try? etc.

PR #151, opened 2019-09-06
[MOB-659] - InAppManager.remove(:message) method issues

tapashmajumder:
Need to add back method in public api.

### DIFF
--- a/swift-sdk/Internal/InAppManager.swift
+++ b/swift-sdk/Internal/InAppManager.swift
@@ -32,6 +32,7 @@ protocol IterableInAppManagerProtocolInternal: IterableInAppManagerProtocol, InA
 
 class InAppManager: NSObject, IterableInAppManagerProtocolInternal {
     init(apiClient: ApiClientProtocol,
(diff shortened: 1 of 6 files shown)


swift-sdk/Internal/InAppManager.swift
```diff
@@ -134,7 +136,13 @@ class InAppManager: NSObject, IterableInAppManagerProtocolInternal {
         }
     }
     
-    func remove(message: IterableInAppMessage, location: InAppLocation, source: InAppDeleteSource) {
+    func remove(message: IterableInAppMessage) {
```
roninopf: Is this necessary if `func remove(message: IterableInAppMessage, location: InAppLocation = .unknown, source: InAppDeleteSource = .unknown)` exists as a function with defaults?

tapashmajumder: Yes. Because that was part of public API. It will work fine with Swift but will be problematic for Objective C because it doesn't have default parameters.

PR #157, opened 2019-09-10
[MOB-702] clean up

roninopf:


tapashmajumder:
Conditional + 1. LGTM. Feel free to merge after the following
1. Talking to Mahavir, Victor and DT. All three must be in sync.
2. Please update the quip docs accordingly.

PR #161, opened 2019-09-15
Updated for Swift 5 Support

jstngoulet:
- This project has been updated to support swift 5 for both the 'AppExtensions' framework and the 'SDK' framework. The code base was migrated to Swift 5 along with an update to the podspecs and build settings.

- Also in this update includes a small update to the 'DeviceInfo.swift' and 'IterableAppIntegrationInternal.swift' classes that add in the new '@unknown default' case for Swift switch statements.

- An unmodified var in the 'CommonExtensions.swift' class showed a compiler warning, so that value was updated to be a constant, using the 'let' declaration.


tapashmajumder:
Hi. We are going to release a new version with Swift 5 support soon. It is already in the works.

PR #165, opened 2019-09-19
[MOB-709] - IterableAPI public API remove deprecated etc

tapashmajumder:


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -338,11 +338,10 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
                                  forResult: apiClient.inAppConsume(messageId: messageId))
     }
     
-    func inAppConsume(message: IterableInAppMessage, location: InAppLocation, source: InAppDeleteSource) {
+    func inAppConsume(message: IterableInAppMessage, location: InAppLocation = .unknown, source: InAppDeleteSource = .unknown) {
```
davidtruong: Does unknown actually get passed to the backend?

Currently the backend only expects in-app or inbox, so the api request might fail on this. https://github.com/Iterable/Iterable/blob/master/app/models/api/InboxMessageContext.scala#L44
Is there a use case for not passing in this location?

If we want to accept no value for the location, then we will need to change the backend to support a non-existent value.

Tests/swift-sdk-swift-tests/DeeplinkTests.swift
```diff
@@ -35,7 +35,9 @@ class DeeplinkTests: XCTestCase {
         
         setupRedirectStubResponse(location: redirectLocation, campaignId: campaignId, templateId: templateId, messageId: messageId)
         
-        IterableAPI.getAndTrack(deeplink: URL(string: iterableRewriteURL)!) { redirectUrl in
+        IterableAPI.handle(universalLink: URL(string: iterableRewriteURL)!)
```
davidtruong: Shouldn't handle be inside of getAndTrack? Since handle should use the redirectUrl.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: No unknown does not get passed to the backend. If the value is unknown we don't sent any value to the backend.

Tests/swift-sdk-swift-tests/DeeplinkTests.swift
tapashmajumder: `handle`... and `getAndTrack..` are two different methods doing the same thing.
What handle does is it gets the link inside the deep link from the server and calls `IterableURLDelegate` on this. 
What getAndTrack does is it gets the link inside the deep link and then handles the link itself.

That is the reason we are deprecating `getAndTrack` and asking users to move to `handle` instead

swift-sdk/Internal/IterableAPIInternal.swift
davidtruong: In the api contract location is required, so the api call would fail if the location value is not there. Sounds like we would need a change to the api to support this.

tapashmajumder: We will not always know the location. I can pass a default location but I think that is not ideal.

tapashmajumder: Plus previous versions of SDK will not have location. So that needs to be optional anyway.

davidtruong: Technically it adheres to backwards since the messageContext object is optional. 
I'll go ahead an make location optional as well.

tapashmajumder: Ah. I see.

tapashmajumder: In that case.... I am also OK providing a default location.

tapashmajumder: I have created a separate Jira for "unknown/other". https://iterable.atlassian.net/browse/MOB-748

PR #168, opened 2019-09-23
[MOB-748] - rename unknown to other

tapashmajumder:
Also defaults to 'inApp' where it makes sense.

### DIFF
--- a/swift-sdk/Internal/IterableAPIInternal.swift
+++ b/swift-sdk/Internal/IterableAPIInternal.swift
@@ -404,8 +404,8 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
     
     private var launchOptions: [UIApplication.LaunchOptionsKey: Any]?
     
-    private lazy var apiClient: ApiClient = {
-        ApiClient(apiKey: apiKey, authProvider: self, endPoint: .ITBL_ENDPOINT_API, networkSession: networkSession)
+    lazy var apiClient: ApiClient = {
+        ApiClient(apiKey: apiKey, authProvider: self, endPoint: config.apiEndpoint, networkSession: networkSession)
     }()
     
     var networkSession: NetworkSessionProtocol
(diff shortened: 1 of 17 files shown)


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -326,10 +326,7 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
     func track(inAppDelivery message: IterableInAppMessage) {
         IterableAPIInternal.call(successHandler: IterableAPIInternal.defaultOnSucess(identifier: "trackInAppDelivery"),
                                  andFailureHandler: IterableAPIInternal.defaultOnFailure(identifier: "trackInAppDelivery"),
-                                 forResult: apiClient.track(inAppDelivery: message.messageId,
-                                                            saveToInbox: message.saveToInbox,
-                                                            silentInbox: message.saveToInbox && message.trigger.type == .never,
-                                                            deviceMetadata: deviceMetadata))
+                                 forResult: apiClient.track(inAppDelivery: InAppMessageContext.from(message: message, location: .other, deviceMetadata: deviceMetadata)))
```
vbabenkoru: Do we even need `location` for `inAppDelivery`? It will always be set to `other`, which sounds confusing since it is not a third option other than in-app or inbox, but rather that we don't know, so it should be empty.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -300,14 +300,14 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
         IterableAPIInternal.call(successHandler: IterableAPIInternal.defaultOnSucess(identifier: "trackInAppClick"),
                                  andFailureHandler: IterableAPIInternal.defaultOnFailure(identifier: "trackInAppClick"),
-                                 forResult: apiClient.track(inAppClick: InAppMessageContext(message: message, location: location, deviceMetadata: deviceMetadata), clickedUrl: clickedUrl))
+                                 forResult: apiClient.track(inAppClick: InAppMessageContext.from(message: message, location: location, deviceMetadata: deviceMetadata), clickedUrl: clickedUrl))
     }
     
-    func trackInAppClose(_ message: IterableInAppMessage, location: InAppLocation = .unknown, source: InAppCloseSource = .unknown, clickedUrl: String? = nil) {
-        let result = apiClient.track(inAppClose: InAppMessageContext(message: message, location: location, deviceMetadata: deviceMetadata),
+    func trackInAppClose(_ message: IterableInAppMessage, location: InAppLocation = .inApp, source: InAppCloseSource = .other, clickedUrl: String? = nil) {
```
vbabenkoru: Can we make these parameters required and remove `InAppCloseSource.other`? How would a developer use `InAppCloseSource.other`?

swift-sdk/Internal/IterableAPIInternal.swift
vbabenkoru: Actually the backend doesn't even use this parameter, `InAppDelivery` doesn't have `location` in its context.

PR #175, opened 2019-09-26
[MOB-793] - Prepend viewport meta tag.

tapashmajumder:
I have verified that this fix works for templates with bee editor and otherwise.


tapashmajumder:
@vbabenkoru  We are going to try and fix this on the backend. Closing this PR for now.

PR #180, opened 2019-10-02
[MOB-821] deselect inbox row

roninopf:


swift-sdk/IterableInboxViewController.swift
```diff
@@ -123,7 +123,9 @@ open class IterableInboxViewController: UITableViewController {
     
     // MARK: - UITableViewDelegate (Optional Functions)
     
-    open override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
+    open override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        
         let message = viewModel.message(atRow: indexPath.row)
```
tapashmajumder: This appears to be fine. Just curious what problem are we seeing if we don't add this line?

PR #188, opened 2019-10-08
[MOB-845] Test EmptyInAppManager

roninopf:


roninopf:
> +1.
> Although I don't think this class needs any tests.

Yeah, we aren't expecting to change it from as is, but it's good to have the coverage on the reporting side.

PR #189, opened 2019-10-08
[MOB-843] Add a check for attribution description

roninopf:


roninopf:
> +1
> `description` is something that should not affect how our code works. So although this increases our coverage it doesn't do anything about increasing our code reliability. Actually I am not even sure what this description field is doing. Maybe we should look into it in the future and remove it if it is not being used.

That's definitely true, and it's useful for debugging, basically. Most things in Swift have their own `description` string, and you can override it to add in custom info.

PR #190, opened 2019-10-08
[MOB-856] test deprecated trackInAppOpen

roninopf:


Tests/swift-sdk-swift-tests/TestUtils.swift
```diff
@@ -84,6 +84,22 @@ struct TestUtils {
...
+        if let userId = userId {
+            validateMatch(keyPath: KeyPath(JsonKey.userId), value: userId, inDictionary: body)
+        }
+        
+        let contextKey = "\(JsonKey.inAppMessageContext.jsonKey)"
+        validateMatch(keyPath: KeyPath("\(contextKey).\(JsonKey.saveToInbox.jsonKey)"), value: saveToInbox, inDictionary: body)
+        validateMatch(keyPath: KeyPath("\(contextKey).\(JsonKey.silentInbox.jsonKey)"), value: silentInbox, inDictionary: body)
+    }
+    
     static func validateDeviceInfo(inBody body: [String: Any]) {
```
tapashmajumder: I don't like putting deprecated code here. Can we modify `TestUtils.validateMessageContext` method to take inAppLocation as an optional? That way we would have just one function, not two.

Tests/swift-sdk-swift-tests/DeprecatedFunctionsTests.swift
```diff
@@ -0,0 +1,61 @@
...
+            TestUtils.validateDeviceInfo(inBody: body)
+            
+            expectation1.fulfill()
+        }
+        
+        IterableAPI.track(inAppOpen: message.messageId)
+        wait(for: [expectation1], timeout: testExpectationTimeout)
+    }
+    
+    func testDeprecatedTrackInAppClick() {}
```
tapashmajumder: Please remove empty functions or better yet, implement them :).

Tests/swift-sdk-swift-tests/DeprecatedFunctionsTests.swift
```diff
@@ -0,0 +1,67 @@
+//
+//  DeprecatedFunctionsTests.swift
+//  swift-sdk-swift-tests
+//
```
tapashmajumder: Not needed for +1 but I usually remove the above two lines when I add new files. Sometimes we need to change file names when refactoring and this becomes stale. Besides, this doesn't really add any information. Anyway, this is just a comment for future commits.

PR #209, opened 2019-10-30
[HOLD] [MOB-933] Remove iOS 9 support for future SDKs

roninopf:


tapashmajumder:
Did we get approval for doing this from Suhel or Mahavir? We should get a +1 from them.

roninopf:
> Did we get approval for doing this from Suhel or Mahavir? We should get a +1 from them.

Whoops. You're right, we should. It'll be ready for when they say yes, at least. :P

Tests/common/CommonMocks.swift
```diff
@@ -74,7 +73,6 @@ public class MockCustomActionDelegate: NSObject, IterableCustomActionDelegate {
 @objcMembers
 public class MockUrlOpener: NSObject, UrlOpenerProtocol {
     var ios10OpenedUrl: URL?
```
tapashmajumder: We should probably rename to `openedUrl`

roninopf:
Closing for new future PR to drop iOS 9 support

PR #218, opened 2019-11-13
docs: Clarify documentation for pushIntegrationName config

bricker:
This wasn't made clear in any documentation, thought it could help the next person.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ Iterable API key:
     
 ```swift
 let config = IterableConfig()
-config.pushIntegrationName = "<your-iterable-push-integration-name>"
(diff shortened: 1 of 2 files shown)


roninopf:
Thanks for the PR! We'll get to this soon; as you can tell from time of writing, we've got a LOT going on with PRs.

roninopf:
Hi @bricker , thanks for bringing this to our attention; unfortunately, your changes aren't entirely correct, since it depends on the project itself inside Iterable (older projects, especially, will most likely have different names than the bundle ID), and the proper fix to documenting this is quite a bit bigger, so we've created a ticket which I personally will be making sure we get in sooner rather than later.

To add, you can always find what integration names are being used at this URL (with your Iterable account logged in: https://app.iterable.com/integrations/mobileApps)

PR #224, opened 2019-11-22
[MOB-971] Enable distribution toggle

roninopf:


tapashmajumder:
Please see comments in Jira.

roninopf:
> Thanks! This looks quite good. Can you please make sure that this solution is working for them?

Yes, will do!

roninopf:
Verified with customer that the frameworks generated are working.

PR #225, opened 2019-11-27
[MOB-927] Add inboxSessionId

roninopf:


roninopf:
Locally ran through all the unit tests and they all passed.

tapashmajumder:
Few suggestions/observations
1. We should look into removing inboxSessionId from IterableInAppMessage class and add this to InAppMessageContext. 
2. I don't see anywhere that IterableInAppMessage.inboxSessionId value is being set.
3. We should probably write or change unit tests to make sure that inboxSessionId is being passed for calls other than trackInboxSession.

roninopf:
At this point, just two tasks remain:
- unit tests
- converting the internal use of `inAppConsume` to track `inboxSessionId`

roninopf:
Clarified that the current addition of `inboxSessionId` to `inAppConsume` is fine since it's meant to be used internally, so unit tests remain.

Tests/swift-sdk-swift-tests/InboxImpressionTrackerTests.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  InboxImpressionTrackerTests.swift
+//  swift-sdk-swift-tests
+//
```
tapashmajumder: nit: Please remove these first two lines. They don't really help much and they will be confusing if we change the name of file.

Tests/swift-sdk-swift-tests/InboxImpressionTrackerTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    func testUpdateVisibleRows() {
+        let rowInfo1 = InboxImpressionTracker.RowInfo(messageId: IterableUtil.generateUUID(), silentInbox: false)
+        let rowInfo2 = InboxImpressionTracker.RowInfo(messageId: IterableUtil.generateUUID(), silentInbox: true)
+        let rowInfo3 = InboxImpressionTracker.RowInfo(messageId: IterableUtil.generateUUID(), silentInbox: false)
+        
+        let tracker = InboxImpressionTracker()
+        
+        tracker.updateVisibleRows(visibleRows: [rowInfo1])
+        
+        sleep(30)
```
tapashmajumder: Is the time in seconds or milliseconds? If seconds this is too long for a test.

Tests/swift-sdk-swift-tests/InboxSessionManagerTests.swift
```diff
@@ -0,0 +1,85 @@
+//
+//  InboxSessionManagerTests.swift
+//  swift-sdk-swift-tests
```
tapashmajumder: nit: remove these two lines.

Tests/swift-sdk-swift-tests/RequestCreatorTests.swift
```diff
@@ -40,6 +43,102 @@ class RequestCreatorTests: XCTestCase {
...
+        let messageContext1 = InAppMessageContext.from(message: message, location: .inbox, inboxSessionId: inboxSessionId)
+        let request1 = convertToUrlRequest(createRequestCreator().createTrackInAppOpenRequest(inAppMessageContext: messageContext1))
+        TestUtils.validateHeader(request1, apiKey)
+        TestUtils.validate(request: request1, requestType: .post, apiEndPoint: Endpoint.api, path: Const.Path.trackInAppOpen)
+        
+        let body1 = request1.bodyDict
+        TestUtils.validateMatch(keyPath: KeyPath(.email), value: email, inDictionary: body1)
+        TestUtils.validateMatch(keyPath: KeyPath(.messageId), value: messageId, inDictionary: body1)
+        TestUtils.validateMatch(keyPath: KeyPath(.inboxSessionId), value: inboxSessionId, inDictionary: body1)
+        //test contents of messageContext of body1 here
```
tapashmajumder: Not sure that this comment applies here.

Tests/swift-sdk-swift-tests/RequestCreatorTests.swift
```diff
@@ -40,6 +43,102 @@ class RequestCreatorTests: XCTestCase {
...
+        let messageContext2 = InAppMessageContext.from(message: message, location: .inbox)
+        let request2 = convertToUrlRequest(createRequestCreator().createTrackInAppOpenRequest(inAppMessageContext: messageContext2))
+        TestUtils.validateHeader(request2, apiKey)
+        TestUtils.validate(request: request2, requestType: .post, apiEndPoint: Endpoint.api, path: Const.Path.trackInAppOpen)
+        
+        let body2 = request2.bodyDict
+        TestUtils.validateMatch(keyPath: KeyPath(.email), value: email, inDictionary: body2)
+        TestUtils.validateMatch(keyPath: KeyPath(.messageId), value: messageId, inDictionary: body2)
+        TestUtils.validateNil(keyPath: KeyPath(.inboxSessionId), inDictionary: body2)
+        //test contents of messageContext of body2 here
```
tapashmajumder: Not sure about this comment. Remove?

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -301,9 +302,19 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
+                                 forResult: result)
+    }
+    
+    func trackInAppClick(_ message: IterableInAppMessage, location: InAppLocation = .inApp, inboxSessionId: String?, clickedUrl: String) {
+        let result = apiClient.track(inAppClick: InAppMessageContext.from(message: message, location: location, inboxSessionId: inboxSessionId),
+                                     clickedUrl: clickedUrl)
+        IterableAPIInternal.call(successHandler: IterableAPIInternal.defaultOnSucess(identifier: "trackInAppClick"),
+                                 andFailureHandler: IterableAPIInternal.defaultOnFailure(identifier: "trackInAppClick"),
+                                 forResult: result)
     }
```
tapashmajumder: Since IterableAPIInternal is an internal class and will be used with only Swift and not Objc. We don't need method overloads. Just add the extra parameter to existing method.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -315,6 +326,15 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
     
+    func trackInAppClose(_ message: IterableInAppMessage, location: InAppLocation = .inApp, inboxSessionId: String? = nil, source: InAppCloseSource? = nil, clickedUrl: String? = nil) {
+        let result = apiClient.track(inAppClose: InAppMessageContext.from(message: message, location: location, inboxSessionId: inboxSessionId),
+                                     source: source,
+                                     clickedUrl: clickedUrl)
+        IterableAPIInternal.call(successHandler: IterableAPIInternal.defaultOnSucess(identifier: "trackInAppClose"),
+                                 andFailureHandler: IterableAPIInternal.defaultOnFailure(identifier: "trackInAppClose"),
+                                 forResult: result)
+    }
+
```
tapashmajumder: Just add extra parameter `inboxSessionId` to existing method with default value of nil.

swift-sdk/IterableMessaging.swift
```diff
@@ -43,6 +43,12 @@ import Foundation
     /// - parameter source: The source of deletion `inboxSwipe` or `deleteButton`.`
     @objc(removeMessage:location:source:) func remove(message: IterableInAppMessage, location: InAppLocation, source: InAppDeleteSource)
     
+    /// - parameter message: The message to remove.
+    /// - parameter location: The location from where this message was shown. `inbox` or `inApp`.
+    /// - parameter source: The source of deletion `inboxSwipe` or `deleteButton`.`
+    /// - parameter inboxSessionId: The ID of the inbox session that the message originates from.
+    @objc(removeMessage:location:source:inboxSessionId:) func remove(message: IterableInAppMessage, location: InAppLocation, source: InAppDeleteSource, inboxSessionId: String?)
+
```
tapashmajumder: Not needed for this PR. But I think we should take another look at public methods and classes. I think we don't need to expose inboxSessionId publicly. Since this will never be set by developers. But that is for another Jira issue and PR.

swift-sdk/IterableMessaging.swift
```diff
@@ -205,6 +211,9 @@ public extension Notification.Name {
 
 /// Encapsulates Inbox Session
 @objcMembers public final class IterableInboxSession: NSObject, Codable {
```
tapashmajumder: Not needed for this PR.... Does this class need to be public?

Tests/swift-sdk-swift-tests/InboxImpressionTrackerTests.swift
roninopf: Misread what the impression minimum is while I wrote this, haha. Will fix. `sleep` uses seconds.

swift-sdk/IterableMessaging.swift
roninopf: Created a ticket and also put this specific function in there, because I agree, this probably does not need to be public.

swift-sdk/IterableMessaging.swift
roninopf: I also added this to be looked at with the evaluation ticket

PR #228, opened 2019-12-04
Podspec support for both Swift 4.2 and 5.0

ejensen:
The CocoaPod pod spec format has a mechanism to [specify multiple version of Swift](https://guides.cocoapods.org/syntax/podspec.html#swift_versions) that a Pod support. This PR adds Swift 4.2 and 5.0 to the `swift_versions` list since the Iterable the SDK is compatible with both.

Commit 9bd6091 addresses a couple compiler warnings (not errors) when compiling with Swift 5

### DIFF
--- a/Iterable-iOS-AppExtensions.podspec
+++ b/Iterable-iOS-AppExtensions.podspec
@@ -79,9 +79,5 @@ Pod::Spec.new do |s|
 
   # s.public_header_files = "Classes/**/*.h"
 
-  s.pod_target_xcconfig = {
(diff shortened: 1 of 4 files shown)


roninopf:
Hi @ejensen , sorry for the extremely late review. When I first saw this, I tried to investigate to what this change exactly does, but I couldn't figure out the whole scope. Given we just released Mobile Inbox which had a lot of overall changes (including an upgrade to Swift 5), is this PR still needed?

ejensen:
No longer needed since the SDK now only supports Swift 5

PR #231, opened 2019-12-05
[MOB-992] - Inbox Filtering

tapashmajumder:


swift-sdk/IterableInboxViewController.swift
```diff
@@ -47,7 +54,8 @@ open class IterableInboxViewController: UITableViewController {
     
     /// By default, messages are sorted chronologically.
     /// This enumeration has sample comparators that can be used by `IterableInboxViewControllerViewDelegate`.
-    public enum Comparator {
+    /// You can create your own comparators which can be functions or closures
+    public enum SampleComparator {
```
roninopf: Should `Sample` be `Default` instead?

PR #232, opened 2019-12-06
[MOB-1001] misc refactors

roninopf:


swift-sdk/Internal/RequestCreator.swift
```diff
@@ -137,36 +138,33 @@ struct RequestCreator {
...
-                    JsonKey.Commerce.total: total]
+            body[JsonKey.dataFields.jsonKey] = dataFields
         }
         
         return .success(.post(createPostRequest(path: Const.Path.trackPurchase, body: body)))
     }
     
     func createTrackPushOpenRequest(_ campaignId: NSNumber, templateId: NSNumber?, messageId: String?, appAlreadyRunning: Bool, dataFields: [AnyHashable: Any]?) -> Result<IterableRequest, IterableError> {
-        var body = [AnyHashable: Any]()
+        var body: [AnyHashable: Any] = [:]
```
tapashmajumder: We should not have to declare the type that can be inferred. The original code is more concise and recommended.

swift-sdk/Internal/IterableUtil.swift
```diff
@@ -78,10 +79,11 @@ enum IterableResult<T, E> {
...
             if let customActionDelegate = customActionDelegate {
                 _ = customActionDelegate.handle(iterableCustomAction: context.action, inContext: context)
+                
                 return true
-            } else {
-                return false
             }
+            
+            return false
         }
```
tapashmajumder: If you want to remove the else portion, I think this would be better with a `guard` statement instead of `if`.

swift-sdk/Internal/IterableUtil.swift
```diff
@@ -78,10 +79,11 @@ enum IterableResult<T, E> {
         return { _ in
             if let customActionDelegate = customActionDelegate {
                 _ = customActionDelegate.handle(iterableCustomAction: context.action, inContext: context)
+
```
tapashmajumder:
```suggestion
guard let customActionDelegate = customActionDelegate else {
    return false
}

_ = customActionDelegate.handle(iterableCustomAction: context.action, inContext: context)
return true
```

PR #241, opened 2020-01-09
[MOB-1005] Add the read flag as part of the API

roninopf:


tapashmajumder:
Looks good. Do we need to add any tests for this?

roninopf:
> Looks good. Do we need to add any tests for this?

Hmm, I think so, actually, let me review the testing code.

PR #249, opened 2020-01-23
[MOB-1113] Fix updateSubscriptions signature and data types

roninopf:


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -383,20 +383,34 @@ final class IterableAPIInternal : NSObject, PushTrackerProtocol {
         }
     }
     
-    func updateSubscriptions(_ emailListIds: [String]?, unsubscribedChannelIds: [String]?, unsubscribedMessageTypeIds: [String]?) {
+    func updateSubscriptions(campaignId: NSNumber?, templateId: NSNumber?, emailListIds: [NSNumber]?, unsubscribedChannelIds: [NSNumber]?, unsubscribedMessageTypeIds: [NSNumber]?, subscribedMessageTypeIds: [NSNumber]?) {
```
vbabenkoru: campaignId and templateId are in the API only for attribution and they're optional, they're not the main parameters for this call. If we decide to add them, they should be the last arguments in this method.

tapashmajumder: +1. Also we should have default values so `campaignId: NSNumber? = nil` etc.

swift-sdk/IterableAPI.swift
```diff
@@ -353,8 +353,8 @@ public final class IterableAPI : NSObject {
      
      - remark: passing in an empty array will clear subscription list, passing in nil will not modify the list
      */
-    @objc(updateSubscriptions:unsubscribedChannelIds:unsubscribedMessageTypeIds:) public static func updateSubscriptions(_ emailListIds: [String]?, unsubscribedChannelIds: [String]?, unsubscribedMessageTypeIds: [String]?) {
-        internalImplementation?.updateSubscriptions(emailListIds, unsubscribedChannelIds: unsubscribedChannelIds, unsubscribedMessageTypeIds: unsubscribedMessageTypeIds)
+    @objc(updateSubscriptions:templateId:emailListIds:unsubscribedChannelIds:unsubscribedMessageTypeIds:subscribedMessageTypeIds:) public static func updateSubscriptions(campaignId: NSNumber?, templateId: NSNumber?, emailListIds: [NSNumber]?, unsubscribedChannelIds: [NSNumber]?, unsubscribedMessageTypeIds: [NSNumber]?, subscribedMessageTypeIds: [NSNumber]?) {
+        internalImplementation?.updateSubscriptions(campaignId: campaignId, templateId: templateId, emailListIds: emailListIds, unsubscribedChannelIds: unsubscribedChannelIds, unsubscribedMessageTypeIds: unsubscribedMessageTypeIds, subscribedMessageTypeIds: subscribedMessageTypeIds)
     }
```
tapashmajumder: The objc naming does not seem right. The first parameter should be `emailIds`

PR #250, opened 2020-01-24
[MOB-1114] update subscriptions on inbox

roninopf:


swift-sdk/Internal/RequestCreator.swift
```diff
@@ -214,6 +219,14 @@ struct RequestCreator {
         
         body.setValue(for: keyValueForCurrentUser.key, value: keyValueForCurrentUser.value)
         
+        if let campaignId = campaignId?.intValue {
```
vbabenkoru: campaignId and templateId are in the API only for attribution and they're optional, they're not the main parameters for this call. If we decide to add them, they should be the last arguments in this method.

PR #261, opened 2020-02-19
[MOB-1194] clean up internal API implementation

roninopf:


swift-sdk/IterableAPI.swift
```diff
@@ -171,8 +171,8 @@ import UIKit
     @objc(updateUser:mergeNestedObjects:onSuccess:onFailure:)
     public static func updateUser(_ dataFields: [AnyHashable: Any],
                                   mergeNestedObjects: Bool,
-                                  onSuccess: OnSuccessHandler? = nil,
-                                  onFailure: OnFailureHandler? = nil) {
```
tapashmajumder: We should keep `= nil` here because then Swift code can call without specifying both success and failure handlers.

swift-sdk/IterableAPI.swift
```diff
@@ -112,7 +112,7 @@ import UIKit
      - onFailure:   OnFailureHandler to invoke if token registration fails
      */
     @objc(registerToken:onSuccess:OnFailure:)
-    public static func register(token: Data, onSuccess: OnSuccessHandler? = nil, onFailure: OnFailureHandler? = nil) {
```
tapashmajumder: We should keep = nil here because then Swift code can call without specifying both success and failure handlers.

PR #262, opened 2020-02-21
[MOB-1217] - Need to resize webview on view did layout subviews.

tapashmajumder:


davidtruong:
@tapashmajumder: Can you make sure to test this on all in-app types to make sure it still functions properly?

tapashmajumder:
@davidtruong Sure thing!

tapashmajumder:
Verified that the solution works for all in-app types.

PR #263, opened 2020-02-23
[MOB-1220] - reorg mappers

tapashmajumder:
Removed sample mappers from public SDK.


tapashmajumder:
Yes. I will make that change separately. That app depends on Swiftsdk so this needs to be checked in first.

PR #264, opened 2020-02-24
[MOB-1032] update docs

roninopf:


CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 6.2.0
+#### Added
```
bradumbaugh: Suggestion:

>>>
#### Updates
- Moved Mobile Inbox support to GA (no longer in beta), and:
    - Added support for various ways to customize the default interface for a mobile inbox
    - Added a sample project that demonstrates how to customize the default interface for a mobile inbox
    - Added tracking for inbox sessions (when the inbox is visible in the app) and inbox message impressions (when a individual message's item is visible in the mobile inbox message list)
- Added support for Swift Package Manager

PR #271, opened 2020-02-28
[MOB-1247] fix ClassExtensions.swift collisions

roninopf:


Iterable-iOS-SDK.podspec
```diff
@@ -76,7 +76,7 @@ Pod::Spec.new do |s|
 
   s.source_files = "swift-sdk/**/*.{h,m,swift}"
 
-  s.resources = ['swift-sdk/Resources/**/*.{storyboard,xib,xcassets}']
```
tapashmajumder: We should create a separate PR for podspec changes. Also need to test thoroughly after making this change.

PR #277, opened 2020-03-03
[MOB-1250] parallelize tests

roninopf:


swift-sdk.xcodeproj/project.pbxproj
```diff
@@ -867,6 +876,7 @@
 				AC98294A20D9D65E00796DAA /* Foundation.framework */,
 			);
 			name = Frameworks;
+			path = ..;
```
tapashmajumder: Why is this needed?

swift-sdk.xcodeproj/project.pbxproj
```diff
@@ -849,7 +858,7 @@
 			path = "host-app";
 			sourceTree = "<group>";
 		};
-		ACFCA72920EB02DB00BFB277 /* Tests */ = {
```
tapashmajumder: We will need to change `Tests` to `tests` codecov.yml where we ignore this directory.

swift-sdk.xcodeproj/project.pbxproj
roninopf: This was to put `Frameworks` and `Products` under the `misc` folder, just for slightly better organization.

vbabenkoru:
Looks like the test aren't completing faster than without parallelization?

PR #284, opened 2020-04-07
MOB-1367] - Do not use IterableAPI.xxx static methods in InAppTests.

tapashmajumder:


roninopf:
Hmm, I'm getting an error in `testAutoShowInAppMultipleOverride()`, saying `Thread 1: Exception: "API violation - multiple calls made to -[XCTestExpectation fulfill] for track in app open."`

tapashmajumder:
The error message is misleading. Had to change some other tests.

PR #285, opened 2020-04-07
[MOB-1371] rename Const to C

roninopf:


tapashmajumder:
Const is a meaningful name and the intent is clear. Not sure why we are renaming Const to C. I think we should close this PR.

roninopf:
I feel like it's actually clearer, and syntactically easier to manage. My inspiration was from Android like in this example code: https://developer.android.com/guide/topics/resources/providing-resources#fromCodeUses

tapashmajumder:
On the other hand iOS encourages meaningful names even if they are long. I don’t think C is more descriptive than Const.

Ayyanchira:
uhm.. that android example might be misleading because that`R` is kind of an exceptional case. It's a part of Android to link the code with Resources like strings, layouts etc.

vbabenkoru:
Yeah, I agree.. `R` in Android is auto-generated from resources, and is not defined manually. iOS doesn't have a similar convention, so it may be less clear.

roninopf:
Haha alright, will close.

PR #289, opened 2020-04-09
[MOB-1391] fix typo in updateSubscriptions

roninopf:


tapashmajumder:
This could be a breaking change for customers using our Objc API.

PR #293, opened 2020-04-22
[MOB-1377] video test

roninopf:


tapashmajumder:
I don't think we can merge this fix into master. This video file will confuse users of our SDK.

roninopf:
Hmm. How about putting it under `tests` but still linking to it through GitHub like before?

roninopf:
Confirmed that's what we'll do for now

PR #294, opened 2020-04-22
[MOB-1362] in-app presentation timer

roninopf:


swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,48 @@
...
+    init(topViewController: UIViewController, htmlMessageViewController: IterableHtmlMessageViewController) {
+        topVC = topViewController
+        htmlMessageVC = htmlMessageViewController
+        
+        htmlMessageVC.presenter = self
+    }
+    
+    func show() {
+        if #available(iOS 10.0, *) {
+            delayTimer = Timer(timeInterval: 0.15, repeats: false) { _ in
```
tapashmajumder, line 22: Is 0.15 the ideal wait time for us? Did you test different values for this setting?

tapashmajumder:
There may be a memory leak. InAppPresenter has a reference to HtmlMessageVC and HtmlMessageVC has a reference to InAppPresenter.

Can you please confirm that the view controller etc is getting deallocated?

swift-sdk/Internal/InAppPresenter.swift
roninopf: We discussed on using 0.15 since it's around the edge of perception of what would be "too late". It can be tuned further if needed, but it's felt fine (e.g. imperceptible) in my testing thus far.

roninopf:
> Have you tested how this works with slow loading images?

Yes! I'm able to trigger the delay timer with an intentionally huge in-app (in our own test project). I also have a very small text only in-app to test for HTML loading first. Confirmed, they trigger the correct functions as expected in my own wireless network.

tapashmajumder:
InAppParsingTests.testDoNotShowMultipleTimes is failing because of fixes. 
The other failures are unrelated.

Also have you checked that `deInit` of IterableHtmlMessageViewController is being called?

roninopf:
> InAppParsingTests.testDoNotShowMultipleTimes is failing because of fixes.
> The other failures are unrelated.
> 
> Also have you checked that `deInit` of IterableHtmlMessageViewController is being called?

Yep, I think I know the cause of the failed test.

Yep, the `deinit` is definitely being called.

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,62 @@
...
+            present()
+        }
+    }
+    
+    func cancelTimer() {
+        if delayTimer != nil {
+            delayTimer?.invalidate()
+            delayTimer = nil
+            
+            present()
```
tapashmajumder, line 51: Why are we presenting here?

tapashmajumder: I see why. Maybe a better name? Not sure what it should be called.

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,62 @@
...
+        
+        htmlMessageVC.presenter = self
+    }
+    
+    func show() {
+        InAppPresenter.isPresenting = true
+        
+        if #available(iOS 10.0, *) {
+            DispatchQueue.main.async {
+                self.delayTimer = Timer.scheduledTimer(withTimeInterval: self.delayInterval, repeats: false) { _ in
```
tapashmajumder, line 33: Use weak self to avoid memory leak.
`
self.delayTimer = Timer.sche....) { [weak self] in 
    self?.delayTimer = nil
    self?.present()
}
`

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,63 @@
...
+    }
+    
+    func show() {
+        InAppPresenter.isPresenting = true
+        
+        if #available(iOS 10.0, *) {
+            DispatchQueue.main.async {
+                self.delayTimer = Timer.scheduledTimer(withTimeInterval: self.delayInterval, repeats: false) { [weak self] _ in
+                    self?.delayTimer = nil
+                    print("jay TIMER CALLBACK")
```
tapashmajumder, line 35: I think this was left here by mistake.

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+
+import UIKit
+
+class InAppPresenter {
+    static var isPresenting = false
+    
+    private let delayInterval: TimeInterval = 1.0
+    
+    private var topVC: UIViewController
```
tapashmajumder, line 13:
```suggestion
    private var topViewController: UIViewController
```

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,70 @@
...
+    private var htmlMessageVC: IterableHtmlMessageViewController
```
tapashmajumder, line 14:
```suggestion
    private var htmlMessageViewController: IterableHtmlMessageViewController
```

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,70 @@
...
+    private var delayTimer: Timer?
+    
+    init(topViewController: UIViewController, htmlMessageViewController: IterableHtmlMessageViewController) {
+        ITBInfo()
+        topVC = topViewController
```
tapashmajumder, line 19: I don't think we need to abbreviate class members. We should use `self.topViewController = topViewController`

swift-sdk/Internal/InAppPresenter.swift
```diff
@@ -0,0 +1,70 @@
...
+        htmlMessageVC = htmlMessageViewController
+
```
tapashmajumder, line 21: Suggestion. Rename to `htmlMessageViewController`.

tapashmajumder:
I think this is ready to get a +1. We should wait to merge this after we release version 6.2.5 of our SDK.

PR #297, opened 2020-04-29
[MOB-1507] - Change campaignId  in IterableInAppMessage to Int from String

tapashmajumder:


nkotula:
Thank you all for the quick turnaround on this!

I know this has already been reviewed and accepted, but I thought I'd put my 2 cents in.  The discussion on the Android side of things led us to come to the conclusion that the `campaignId` should be nullable/optional since proofs don't contain that value in the JSON.  In Java, it was fairly simple to make the public getter return a wrapped `Long` which could be `null`, indicating that there was no `campaignId` in the message.

I know this needs to work with Objective-C, and the only way I know to make a primitive optional is to wrap it with `NSNumber` which is kind of a pain...  Would it be possible to make that work?  The only thing that makes me nervous is that consumers of the SDK would have to know that `0` is an invalid campaign ID.  It has the feeling of a "magic number."

I'm not the one maintaining this SDK, however, so this is just a discussion point

tapashmajumder:
@nkotula  You have a valid point. I will look into making this an optional NSNumber.

roninopf:
Verified to be working, with the change to optional.

PR #318, opened 2020-05-11
[MOB-1248] notification service extension - fix missing action buttons

roninopf:


notification-extension/ITBNotificationServiceExtension.swift
```diff
@@ -27,9 +27,11 @@ import UserNotifications
         var contentHandlerCalled = false
         contentHandlerCalled = loadAttachment(itblDictionary: itblDictionary)
         
-        if !contentHandlerCalled {
-            if let bestAttemptContent = bestAttemptContent {
-                contentHandler(bestAttemptContent)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+            if !contentHandlerCalled {
```
tapashmajumder, line 31: I think we should add `DispatchQueue.main...` here instead of before the check. 
If content handler is not called the function will return immediately so we will trigger the wait. Otherwise the wait is not necessary.
Please validate that this works before making this change.

PR #321, opened 2020-05-13
[MOB-1605] make more classes actually internal

roninopf:


swift-sdk/Internal/Dwifft/Dwifft.swift
```diff
@@ -308,18 +308,17 @@ fileprivate struct MemoizedSequenceComparison<T: Equatable> {
 
 
 // MARK: - Deprecated
```
tapashmajumder, line 310: We should either remove it or keep it. Commented out code does not look very nice.

PR #325, opened 2020-05-20
[MOB-1519] inapppresenter tests (part 1)

roninopf:


tests/swift-sdk-swift-tests/InAppPresenterTests.swift
```diff
@@ -0,0 +1,69 @@
...
+//
+
+import UIKit
+import WebKit
+import XCTest
+
+@testable import IterableSDK
+
+class InAppPresenterTests: XCTestCase {
+    static let delayTimerInterval = 0.75
```
tapashmajumder, line 13: We should make this parameterized. So that we can test with smaller/larger values.

tapashmajumder:
I think InAppPresenter needs further discussion.

PR #329, opened 2020-06-24
refactor

roninopf:


swift-sdk/Internal/ApiClient.swift
```diff
@@ -91,6 +91,64 @@ class ApiClient: ApiClientProtocol {
         self.deviceMetadata = deviceMetadata
     }
     
+    func getInAppMessages(_ count: NSNumber) -> Future<SendRequestValue, SendRequestError> {
+        return send(iterableRequestResult: createRequestCreator().createGetInAppMessagesRequest(count))
+    }
+    
+    func disableDevice(forAllUsers allUsers: Bool, hexToken: String) -> Future<SendRequestValue, SendRequestError> {
+        return send(iterableRequestResult: createRequestCreator().createDisableDeviceRequest(forAllUsers: allUsers, hexToken: hexToken))
+    }
```
tapashmajumder, line 100: These are also making api requests.

swift-sdk/Internal/ApiClient.swift
```diff
@@ -186,55 +230,17 @@ class ApiClient: ApiClientProtocol {
...
-    private func createRequestCreator() -> RequestCreator {
-        guard let authProvider = authProvider else {
-            fatalError("authProvider is missing")
-        }
-        
-        return RequestCreator(apiKey: apiKey, auth: authProvider.auth, deviceMetadata: deviceMetadata)
+}
+
+// DEPRECATED
+extension ApiClient {
```
tapashmajumder, line 236:
```suggestion
extension ApiClient: ApiClientProtocol {
```
This way the intention is clear.

swift-sdk/IterableAPI.swift
```diff
@@ -86,6 +88,66 @@ import UIKit
...
+    
+    // MARK: Private and Internal
+    
+    static var internalImplementation: IterableAPIInternal?
+    
+    private override init() { super.init() }
+}
+
+// API REQUEST CALLS
+extension IterableAPI {
```
tapashmajumder, line 150: I am not sure I like this. We have pretty much moved the entire implementation to an extension. That would be good if in the future we want to extract an interface/protocol.

I do like the separation into `deprecated` extension though.

swift-sdk/Internal/ApiClient.swift
roninopf: Looks like this isn't possible, I get "Redundant conformance of 'ApiClient' to protocol 'ApiClientProtocol'", otherwise I would. That's a new one.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -92,33 +92,6 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         self.dependencyContainer.createInAppManager(config: self.config, apiClient: self.apiClient, deviceMetadata: deviceMetadata)
```
tapashmajumder, line 92: I am not sure I like moving all functionality into an extension. 

The `deprecated` extension is fine.

swift-sdk/IterableAPI.swift
roninopf: I agree it looks a little weird, but I think the separation is still very helpful, if only because it's easier to mentally model what's an API request (and of course there are a lot of those) and what's SDK specific functionality; previously, the way it was laid out made it difficult to make that distinction. Who knows what our exact future plans are for SDK functionality; that section will almost certainly grow on its own with new features. So that's why I'd like to be very clear in the separation despite the odd sizes.

swift-sdk/Internal/IterableAPIInternal.swift
roninopf: Same response as here: https://github.com/Iterable/swift-sdk/pull/329#discussion_r445080195

swift-sdk/IterableAPI.swift
tapashmajumder: Frankly, I have not seen this type of design elsewhere. It is kind of a code smell to me. Usually extensions are there for protocol conformance or to add functionality to a class. But in this case pretty much the entire class is the extension.

tapashmajumder: See comments about class extensions here
https://docs.swift.org/swift-book/LanguageGuide/Extensions.html

roninopf: Hmm, how else could this be structured? I couldn't think of another way to better separate out sections within the same class. I agree that it's not protocol conformance or new functionality, but I still think this separation is very useful.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -92,6 +92,69 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
+                                               value: Const.UserDefaults.payloadExpiration,
+                                               to: dateProvider.currentDate)
+        localStorage.save(payload: payload, withExpiration: expiration)
+        
+        if let metadata = IterablePushNotificationMetadata.metadata(fromLaunchOptions: payload) {
+            if let templateId = metadata.templateId, let messageId = metadata.messageId {
+                attributionInfo = IterableAttributionInfo(campaignId: metadata.campaignId, templateId: templateId, messageId: messageId)
+            }
+        }
+    }
```
tapashmajumder, line 127: I think we should move these two functions to the private section below.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -92,6 +92,69 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
+    @discardableResult private static func call(successHandler onSuccess: OnSuccessHandler? = nil,
+                                                andFailureHandler onFailure: OnFailureHandler? = nil,
+                                                forResult result: Future<SendRequestValue, SendRequestError>) -> Future<SendRequestValue, SendRequestError> {
+        result.onSuccess { json in
+            onSuccess?(json)
+        }.onError { error in
+            onFailure?(error.reason, error.data)
+        }
+        return result
+    }
```
tapashmajumder, line 156: Move to private section.

swift-sdk/Internal/RequestCreator.swift
```diff
@@ -32,6 +32,45 @@ struct RequestCreator {
...
+        case .pad:
+            return JsonValue.DeviceIdiom.pad
+        case .tv:
+            return JsonValue.DeviceIdiom.tv
+        case .carPlay:
+            return JsonValue.DeviceIdiom.carPlay
+        default:
+            return JsonValue.DeviceIdiom.unspecified
+        }
+    }
```
tapashmajumder, line 70: Create a separate `// MARK: - Private` section and move these functions there.

swift-sdk/IterableAPI.swift
```diff
@@ -46,45 +46,7 @@ import UIKit
...
-     Attribution info (campaignId, messageId etc.) for last push open or app link click from an email.
-     */
-    public static var attributionInfo: IterableAttributionInfo? {
-        get {
-            return internalImplementation?.attributionInfo
-        } set {
-            internalImplementation?.attributionInfo = newValue
-        }
-    }
+    // MARK: - API Request Calls
```
tapashmajumder, line 50: These properties should be at top. So the class definition should usually be like

Class {
    public property
    public property
    public methods
}

PR #331, opened 2020-06-29
[MOB-1775] add way for auth token to be included

roninopf:
This PR adds a way to let our customers use authentication tokens (like [JWT](https://jwt.io/)) be used with our SDK. This is done by:
- adding a new `setEmail`/`setUserId` function that not only does the same thing as `IterableAPI.email = "email address"`/`IterableAPI.userId = "user ID"` but also adds an optional field to set the user's token (which our customers will generate/handle). This does mean that calling the function is the only way to set the token. I made this choice to avoid possible race conditions with setting properties, and because I think it's a better formalized way to use our SDK.
- adding the token to API calls if it's present

This PR also has some minor refactoring just for organizational purposes.

### DIFF
--- a/swift-sdk/Internal/ApiClient.swift
+++ b/swift-sdk/Internal/ApiClient.swift
@@ -57,33 +57,12 @@ protocol ApiClientProtocol: AnyObject {
     func disableDevice(forAllUsers allUsers: Bool, hexToken: String) -> Future<SendRequestValue, SendRequestError>
 }
 
-struct Auth {
-    let userId: String?
-    let email: String?
-    
-    var emailOrUserId: EmailOrUserId {
-        if let email = email {
-            return .email(email)
-        } else if let userId = userId {
-            return .userId(userId)
-        } else {
-            return .none
-        }
-    }
-    
-    enum EmailOrUserId {
-        case email(String)
-        case userId(String)
-        case none
-    }
-}
-
-protocol AuthProvider: AnyObject {
-    var auth: Auth { get }
-}
-
 class ApiClient {
-    init(apiKey: String, authProvider: AuthProvider, endPoint: String, networkSession: NetworkSessionProtocol, deviceMetadata: DeviceMetadata) {
+    init(apiKey: String,
+         authProvider: AuthProvider,
+         endPoint: String,
+         networkSession: NetworkSessionProtocol,
+         deviceMetadata: DeviceMetadata) {
         self.apiKey = apiKey
         self.authProvider = authProvider
         self.endPoint = endPoint
@@ -126,10 +105,16 @@ class ApiClient {
     }
     
     private func createIterableHeaders() -> [String: String] {
-        return [JsonKey.contentType.jsonKey: JsonValue.applicationJson.jsonStringValue,
-                JsonKey.Header.sdkPlatform: JsonValue.iOS.jsonStringValue,
-                JsonKey.Header.sdkVersion: IterableAPI.sdkVersion,
-                JsonKey.Header.apiKey: apiKey]
+        var headers = [JsonKey.contentType.jsonKey: JsonValue.applicationJson.jsonStringValue,
+                       JsonKey.Header.sdkPlatform: JsonValue.iOS.jsonStringValue,
+                       JsonKey.Header.sdkVersion: IterableAPI.sdkVersion,
+                       JsonKey.Header.apiKey: apiKey]
+        
+        if let authToken = authProvider?.auth.authToken {
+            headers[JsonKey.Header.authorization] = "Bearer \(authToken)"
+        }
+        
+        return headers
     }
     
     private let apiKey: String
@@ -139,7 +124,8 @@ class ApiClient {
     private let deviceMetadata: DeviceMetadata
 }
 
-// API REQUEST CALLS
+// MARK: - API REQUEST CALLS
+
 extension ApiClient: ApiClientProtocol {
     func register(hexToken: String,
                   appName: String,
@@ -232,7 +218,8 @@ extension ApiClient: ApiClientProtocol {
     }
 }
 
-// DEPRECATED
+// MARK: - DEPRECATED
+
 extension ApiClient {
     // deprecated - will be removed in version 6.3.x or above
     func track(inAppOpen messageId: String) -> Future<SendRequestValue, SendRequestError> {
--- a/swift-sdk/Internal/ApiClient.swift
+++ b/swift-sdk/Internal/ApiClient.swift
@@ -118,7 +118,7 @@ class ApiClient {
     }
     
     private let apiKey: String
-    private weak var authProvider: AuthProvider?
+    private let authProvider: AuthProvider?
     private let endPoint: String
     private let networkSession: NetworkSessionProtocol
     private let deviceMetadata: DeviceMetadata
--- a/swift-sdk/Internal/ApiClient.swift
+++ b/swift-sdk/Internal/ApiClient.swift
@@ -118,7 +118,7 @@ class ApiClient {
     }
     
     private let apiKey: String
-    private let authProvider: AuthProvider?
+    private weak var authProvider: AuthProvider?
     private let endPoint: String
     private let networkSession: NetworkSessionProtocol
     private let deviceMetadata: DeviceMetadata
--- a/swift-sdk/Internal/Auth.swift
+++ b/swift-sdk/Internal/Auth.swift
@@ -12,7 +12,7 @@ protocol AuthProvider: AnyObject {
 struct Auth {
     let userId: String?
     let email: String?
-    var authToken: String?
+    let authToken: String?
     
     var emailOrUserId: EmailOrUserId {
         if let email = email {
--- a/swift-sdk/Internal/Auth.swift
+++ b/swift-sdk/Internal/Auth.swift
@@ -1,0 +1,32 @@
+//
+//  Created by Jay Kim on 6/25/20.
+//  Copyright © 2020 Iterable. All rights reserved.
+//
+
+import Foundation
+
+protocol AuthProvider: AnyObject {
+    var auth: Auth { get }
+}
+
+struct Auth {
+    let userId: String?
+    let email: String?
+    var authToken: String?
+    
+    var emailOrUserId: EmailOrUserId {
+        if let email = email {
+            return .email(email)
+        } else if let userId = userId {
+            return .userId(userId)
+        } else {
+            return .none
+        }
+    }
+    
+    enum EmailOrUserId {
+        case email(String)
+        case userId(String)
+        case none
+    }
+}
--- a/swift-sdk/Internal/IterableAPIInternal.swift
+++ b/swift-sdk/Internal/IterableAPIInternal.swift
@@ -136,13 +136,13 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         }
     }
     
-    func setEmail(_ email: String, _ token: String? = nil) {
+    func setEmail(_ email: String, withToken token: String? = nil) {
         authToken = token
         
         self.email = email
     }
     
-    func setUserId(_ userId: String, _ token: String? = nil) {
+    func setUserId(_ userId: String, withToken token: String? = nil) {
         authToken = token
         
         self.userId = userId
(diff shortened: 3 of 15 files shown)


swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -600,7 +630,7 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         }
     }
     
-    private func updateSDKVersion() {
```
tapashmajumder, line 603: I like `SDK` better. We should usually try to capitalize 3 or less letter abbreviations.

swift-sdk/IterableAPI.swift
```diff
@@ -114,6 +114,34 @@ import UIKit
...
+    
+    /// Set the user of the SDK instance to the email address specified
+    ///
+    /// - parameter email: the email of the user for the SDK instance
+    /// - parameter token: the associated authentication token for the user
+    @objc(setEmail:token:)
+    public static func setEmail(_ email: String, _ token: String? = nil) {
+        internalImplementation?.setEmail(email, token)
+    }
+
```
tapashmajumder, line 130: We also need `setEmail:` method. We can't break any public API. Objc code that does not have token will not work with this. Swift will continue to work because token is optional parameter.

tapashmajumder: Or do we already have the old method? In that case you can ignore this comment..

roninopf: You're correct, we don't have just `setEmail`/`setUserId`, I'll add that.

swift-sdk/Internal/IterableAPIInternal.swift
roninopf: Relegated to future discussion

swift-sdk/IterableAPI.swift
roninopf: I was mistaken, those functions are already generated on the Objective-C side, so we're fine!

swift-sdk/Internal/Auth.swift
```diff
@@ -0,0 +1,32 @@
...
+import Foundation
+
+protocol AuthProvider: AnyObject {
+    var auth: Auth { get }
+}
+
+struct Auth {
+    let userId: String?
+    let email: String?
+    var authToken: String?
```
tapashmajumder, line 15: change to `let`. Unless there is a good reason, we should not create mutable structs.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -134,6 +136,26 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         }
     }
     
+    func getEmail() -> String? {
+        return _email
+    }
```
tapashmajumder, line 141: There is already `email` property. We don't need this method.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -134,6 +136,26 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
+    
+    func setEmail(_ email: String, _ token: String? = nil) {
+        authToken = token
+        
+        self.email = email
+    }
+    
+    func getUserId() -> String? {
+        return _userId
+    }
```
tapashmajumder, line 151: There is already `userId` property. We don't need this method.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -28,7 +28,8 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
             
             _email = newValue
             _userId = nil
-            storeEmailAndUserId()
+
```
tapashmajumder, line 31: `authorization = nil` for symmetry

tapashmajumder: I mean `authToken = nil`

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -46,7 +47,8 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
             
             _userId = newValue
             _email = nil
-            storeEmailAndUserId()
+
```
tapashmajumder, line 50: add `authToken = nil` here

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -384,11 +412,14 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
             disableDeviceForCurrentUser()
         }
         
+        authToken = nil
```
tapashmajumder, line 415: No need to do this just for authToken. You can remove this line. We should do symmetric things for `userId`, `email` and `authToken`.

swift-sdk/IterableAPI.swift
```diff
@@ -114,6 +114,34 @@ import UIKit
         internalImplementation?.removeDeviceAttribute(name: name)
     }
     
+    /// Get the email address of the user of the SDK instance
+    public static func getEmail() -> String? {
+        return internalImplementation?.getEmail()
+    }
+
```
tapashmajumder, line 121: We don't need this method. We already have `IterableAPI.email` getter.

swift-sdk/IterableAPI.swift
```diff
@@ -114,6 +114,34 @@ import UIKit
...
+    @objc(setEmail:token:)
+    public static func setEmail(_ email: String, _ token: String? = nil) {
+        internalImplementation?.setEmail(email, token)
+    }
+    
+    /// Get the user ID of the user of the SDK instance
+    public static func getUserId() -> String? {
+        return internalImplementation?.getUserId()
+    }
+
```
tapashmajumder, line 135: We already have `IterableAPI.userId` getter.

tests/swift-sdk-swift-tests/IterableAPITests.swift
```diff
@@ -170,17 +171,42 @@ class IterableAPITests: XCTestCase {
...
-    func testEmailUserIdPersistence() {
+    func testEmailPersistence() {
         let internalAPI = IterableAPIInternal.initializeForTesting()
         
         internalAPI.email = IterableAPITests.email
         XCTAssertEqual(internalAPI.email, IterableAPITests.email)
         XCTAssertNil(internalAPI.userId)
+    }
+    
+    func testEmailUserIdPersistence() {
```
tapashmajumder, line 182: `testUserIdPersistence`

swift-sdk/Internal/IterableAPIInternal.swift
roninopf: I can't do that because `authToken`for the new email gets set before this call.

https://github.com/Iterable/swift-sdk/pull/331/files#diff-a003af0114c3087fec8bdff95b1a37c7R139

swift-sdk/Internal/ApiClient.swift
```diff
@@ -126,20 +105,27 @@ class ApiClient {
...
+        
+        if let authToken = authProvider?.auth.authToken {
+            headers[JsonKey.Header.authorization] = "Bearer \(authToken)"
+        }
+        
+        return headers
     }
     
     private let apiKey: String
-    private weak var authProvider: AuthProvider?
```
tapashmajumder, line 136: Why do we need to have a strong reference to authProvider here.

tapashmajumder: ApiClient has a reference to AuthProvider (APIInternal) and AuthProvider (ApiInternal) has a reference to ApiClient.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: I see you are calling IterableAPI.email from IterableAPI.setEmail... OK.

swift-sdk/Internal/IterableAPIInternal.swift
tapashmajumder: Still need to rename it to SDK. I think that is the style used by Apple. Look at their APIs. We should try to follow Apple guidelines as much as possible.

swift-sdk/IterableAPI.swift
```diff
@@ -114,6 +114,24 @@ import UIKit
         internalImplementation?.removeDeviceAttribute(name: name)
     }
     
+    /// Set the user of the SDK instance to the email address specified
+    ///
+    /// - parameter email: the email of the user for the SDK instance
+    /// - parameter token: the associated authentication token for the user
+    @objc(setEmail:token:)
+    public static func setEmail(_ email: String, _ token: String? = nil) {
```
tapashmajumder, line 122: Recommendation: rename to `setEmail(_ email: String, withToken token: String? = nil)`

tapashmajumder: Also objc name to `setEmail:withToken`

swift-sdk/Internal/ApiClient.swift
roninopf: When I was writing the test, it wouldn't succeed without being strongly referenced; somewhere along the way it kept getting lost. I'll take any suggestions on how to fix this while keeping it weakly referenced, but I couldn't think of one at the time.

roninopf: Resolved, problem was in test implementation. Other tests to be fixed at a later date.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -134,6 +136,18 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         }
     }
     
+    func setEmail(_ email: String, _ token: String? = nil) {
```
tapashmajumder, line 139: We should name this the same as in IterableAPI which is `setEmail:withToken`

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -134,6 +136,18 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
+        authToken = token
+        
+        self.email = email
+    }
+    
+    func setUserId(_ userId: String, _ token: String? = nil) {
```
tapashmajumder, line 145: Consider renaming to `setUserId:withToken`

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -134,6 +136,18 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
         }
     }
     
+    func setEmail(_ email: String, _ token: String? = nil) {
+        authToken = token
+        
+        self.email = email
```
tapashmajumder, line 142: It seems that calling `self.email` will set `authToken = nil`.

tapashmajumder: self.email will call logoutPreviousUser which will set authToken to nil.

Please add a test where the email is changed from 'user1@example.com` to `user2@example.com`. That test will fail.

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -20,35 +20,31 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
-            guard newValue != _email else {
-                return
+            if newValue == nil {
+                logoutPreviousUser()
+            } else {
+                guard let newEmail = newValue else {
+                    return
+                }
+                
+                setEmail(newEmail)
```
tapashmajumder, line 30: We should be able to replace this with
`setEmail(newValue)`

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -20,35 +20,31 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
...
-            guard newValue != _userId else {
-                return
+            if newValue == nil {
+                logoutPreviousUser()
+            } else {
+                guard let newUserId = newValue else {
+                    return
+                }
+                
+                setUserId(newUserId)
```
tapashmajumder, line 46: We should be able to replace this with setUserId(newValue)

PR #332, opened 2020-06-30
xcode 12 build fix

akshya672222:


roninopf:
Hi @akshya672222 , thanks for submitting this PR!

Just letting you know, it may be a little bit before we merge it; unfortunately, this change isn't backwards compatible so we'll wait until we're on Xcode 12's release to do so (whenever it's out of beta).

PR #342, opened 2020-07-08
[MOB-1799] refactor

roninopf:


swift-sdk/Internal/IterableActionRunner.swift
```diff
@@ -76,14 +79,12 @@ struct IterableActionRunner {
...
-    // MARK: Private
-    
     private static func shouldOpenUrl(url: URL, from source: IterableActionSource) -> Bool {
         if source == .push || source == .inApp, let scheme = url.scheme, scheme == "http" || scheme == "https" {
             return true
-        } else {
-            return false
         }
+        
+        return false
```
tapashmajumder, line 87: Not sure why this is better than what was existing?

tapashmajumder: In general short circuiting is helpful for methods with long bodies.

roninopf: Oh, that was just a clarity change since there's no point in having the `else` there with just a `return` in it.

tapashmajumder: Usually I also don’t like unnecessary else statements containing code. And short circuiting is good. But in this case this does not feel like short circuiting. I like the symmetry of if else. Let’s leave it as it is.

roninopf: I think it still short circuits the same way in both scenarios

tapashmajumder: Of course. Both will do the same thing. I just like the symmetry of this better.

Consider the two scenarios below. For the first, I like return inside else. For the second scenario I like return outside as you propose.

// scenario 1 (currently existing)
if condition1 && condition2 {
   return true
} else {
  return false
}

// scenario 2
if condition1 && condition 2 {
return true
} else {
  // notice more statements
  statement 1
  statement 2
  return false
}

// Scenario 2 should be changed to 
if condition1 && condition 2 {
  return true
}

statement 1
statement 2
return false
}


If you look at Scala code (not that we should follow Scala guidelines for Swift), that is how they write most of their code (without the explicit return).

swift-sdk/Internal/Auth.swift
```diff
@@ -30,3 +30,9 @@ struct Auth {
         case none
     }
 }
+
+struct DeviceMetadata: Codable {
+    let deviceId: String
+    let platform: String
+    let appPackageName: String
+}
```
tapashmajumder, line 38: I don't think this belongs in Auth.swift file maybe in APIClient.swift file.

PR #350, opened 2020-07-21
[MOB-549] sdk docs

roninopf:


tapashmajumder:
We should replace javadoc style comments /** */ with swift style comments with ///.

All swift libraries are commented this way. This is also recommended here
https://google.github.io/swift/#documentation-comments

roninopf:
> We should replace javadoc style comments /** */ with swift style comments with ///.
> 
> All swift libraries are commented this way. This is also recommended here
> https://google.github.io/swift/#documentation-comments

Interesting, [Apple's own documentation](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/AddingMarkup.html#//apple_ref/doc/uid/TP40016497-CH3-SW1) uses /***/, but I agree with you; I'll change it all to ///.

tapashmajumder:
When you look at actual source code for Swift libraries by selecting a class and then 'Jump to definition' all comments are like this. So I think we are doing the right thing.

roninopf:
> When you look at actual source code for Swift libraries by selecting a class and then 'Jump to definition' all comments are like this. So I think we are doing the right thing.

Yeah, definitely agree. I think the link I used might be out of date, stylistically.

PR #356, opened 2020-07-28
[MOB-1890] - Prevent sending messageId as null

Ayyanchira:


tapashmajumder:
I think it will be better to change ApiClientProtocol to not take messageId with nil values. 
Return an error in `IterableAPIInternal.trackPushOpen` when messageId is nil.
That way all the methods in ApiClientProtocol that take messageId will be consistent.

swift-sdk/Internal/RequestCreator.swift
```diff
@@ -138,6 +138,11 @@ struct RequestCreator {
         var body = [AnyHashable: Any]()
         var reqDataFields = [AnyHashable: Any]()
         
+        guard let messageId = messageId else {
+            ITBError("messageId is nil")
+            return .failure(IterableError.general(description: "messageId is nil"))
+        }
+
```
tapashmajumder, line 145: Change method signature `messageId: String`.

tapashmajumder:
I think we can change IterableAPI.trackPushOpen signature also to take messageId: String.

PR #357, opened 2020-07-29
[MOB-1913] - Make IterableRequest Codable.

tapashmajumder:


swift-sdk/Internal/IterableRequest.swift
```diff
@@ -0,0 +1,94 @@
...
+    enum CodingKeys: String, CodingKey {
+        case type
+        case value
+    }
+    
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let type = try container.decode(String.self, forKey: .type)
+        switch type.lowercased() {
+        case "get":
```
roninopf, line 25: Should "get" and "post" just be constants in our SDK? We could also remove the need to lower case in the switch.

tapashmajumder:
OK. Applied the suggested fix.

PR #362, opened 2020-08-18
[MOB-1974] in app display pause

roninopf:


swift-sdk/Internal/InAppManager.swift
```diff
@@ -35,7 +35,7 @@ protocol IterableInternalInAppManagerProtocol: IterableInAppManagerProtocol, InA
     func remove(message: IterableInAppMessage, location: InAppLocation, source: InAppDeleteSource, inboxSessionId: String?)
 }
 
-class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
+class InAppManager: NSObject {
```
tapashmajumder, line 38: Nit: For discussion: In this case the entire purpose of InAppManager is that it is an implementation of IterableInAppManagerProtocol. I think it may be better to not move the protocol work out of this.

Usually when we have `extension ClassX: protocol PX ` we are implementing a secondary protocol implementation other than the primary purpose of the class.

For example, AppDelegate can be implementing the `UNUserNotificationCenterDelegate`. In that case we will have 
`extension AppDelegate: UNUserNotificationCenterDelegate`

tapashmajumder: Maybe we could put a `// MARK: IterableManagerProtocol implementation.` and put all methods there. 

Also if a protocol has variables you will not be able to put the the variable declaration into the `extension ClassX: protocol PX` section.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -186,6 +144,20 @@ class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
...
+    
+    func setAutoDisplayPaused(_ paused: Bool) {
+        autoDisplayPaused = paused
+        
+        if !autoDisplayPaused {
+            _ = scheduleSync()
+        }
+    }
+    
+    // MARK: - InAppDisplayChecker
```
tapashmajumder, line 159: Not needed for +1. If you want, this one you can move outside using
`extension InAppManager: InAppDisplayChecker`

tapashmajumder:
We need to add one more test
1. When display is paused and then resumed ensure that the message which was shown earlier is now shown.
I think this test will fail.

swift-sdk/Internal/InAppManager.swift
```diff
@@ -613,3 +620,29 @@ extension InAppManager: InAppNotifiable {
         return result
     }
 }
+
+extension InAppManager: InAppDisplayChecker {
+    func isOkToShowNow(message: IterableInAppMessage) -> Bool {
+        guard !isAutoDisplayPaused() else {
```
tapashmajumder, line 626: nit: Not needed for +1. why not call `autoDisplayPaused` variable? 

Style-wise, if `isAuthoDisplayPaused` was a property rather than a variable then it would make sense to call `isAuthoDisplayPaused`  it rather than the function.

tapashmajumder (approved):
+1

PR #365, opened 2020-08-25
[MOB-2026] auth failure delegate

roninopf:


swift-sdk/Internal/NetworkHelper.swift
```diff
@@ -10,10 +10,14 @@ typealias SendRequestValue = [AnyHashable: Any]
 struct SendRequestError: Error {
     let reason: String?
     let data: Data?
+    let responseCode: Int?
```
tapashmajumder, line 13: It is named `httpStatusCode` in offline-events branch.

tapashmajumder:
We have added the variable `responseCode` but I don't see where we are setting it in SendRequestError. Can you please take a look at offline-events branch?

roninopf:
> We have added the variable `responseCode` but I don't see where we are setting it in SendRequestError. Can you please take a look at offline-events branch?

It was used the same way `httpStatusCode` is; but I'm only using it in the 401. But, I'll reconcile the version on the `offline-events` branch into here instead.

swift-sdk/Constants.swift
```diff
@@ -195,6 +195,10 @@ public enum JsonKey: String, JsonKeyRepresentable {
         static let defaultAction = "defaultAction"
     }
     
+    public enum Response {
+        static let code = "code"
```
tapashmajumder, line 199:
```suggestion
        static let iterableCode = "iterableCode"
```

tapashmajumder (approved):
+1

PR #366, opened 2020-08-25
xcode 12 compilation

tsheaff:
Forked off of master, not v6.2.9

This gets the code compiling for Xcode 12 Beta 5 (the latest as of August 24th), which is required to build against iOS 14, which is coming out to customers in just a few weeks


roninopf:
Thanks @tsheaff I've begun reviewing this, and will proceed with a release, ASAP.

tapashmajumder:
> Yes we did see an issue with this in Xcode 12 Beta 5. The new code unwraps the optional one additional time. It was complaining about a `??` double optional.
> 
> Unfortunately, I have not done the work of digging into exactly which change in the swift compiler in xcode 12 caused this, but this change did fix the issue and allow us to compile.
> 
> Are you not getting this same issue with Xcode 12 beta (the latest is now beta 6 as of this moment).

Yes I am not seeing this with Xcode 12 Beta 6. This is a compilation error right? Can you please confirm.

roninopf:
Looks like Xcode 12 beta 6 fixed the exhaustive switch case error into a warning, too.

roninopf:
I checked again with a clean build on beta 6, and I've confirmed it's now only a warning, so I'll be closing this PR.

We do have some broken UI tests on it, but otherwise, just more warnings to clear up.

Thanks for submitting, though, we've adjusted our plan for handling future iOS releases, to give better visibility beforehand.

PR #375, opened 2020-09-11
[MOB-2072] retrieve new token

roninopf:


swift-sdk/AuthManager.swift
```diff
@@ -0,0 +1,84 @@
...
+    
+    // MARK: - Private/Internal
+    
+    private var authToken: String?
+    
+    private var localStorage: LocalStorageProtocol
+    
+    private let onAuthTokenRequestedCallback: (() -> String?)?
+    
+    private static func decodeExpirationDateFromAuthToken(_ authToken: String) -> Int? {
```
tapashmajumder, line 64: I like that this method appears very isolated and has no dependencies. 

Can you please add tests for this?

swift-sdk/Internal/IterableAPIInternal.swift
```diff
@@ -466,13 +464,11 @@ final class IterableAPIInternal: NSObject, PushTrackerProtocol, AuthProvider {
     private func storeAuthData() {
```
tapashmajumder, line 464: Since we are not storing authToken here anymore maybe we should rename this method accordingly? `storeUserCredentials`? (Not sure about this name either).

roninopf: I chose `storeIdentifierData`, I think it's more fitting?

tapashmajumder (approved):
+1

PR #376, opened 2020-09-11
[MOB-2070] JWT expiration

roninopf:


swift-sdk/AuthManager.swift
```diff
@@ -51,16 +56,43 @@ class AuthManager: IterableInternalAuthManagerProtocol {
     
     func retrieveAuthToken() {
         authToken = localStorage.authToken
+        
+        queueAuthTokenExpirationRefresh(authToken)
     }
     
     // MARK: - Private/Internal
     
+    private static let refreshWindow = 60
```
tapashmajumder, line 65: Consider allowing setting this variable in constructor in addition to a default value.

tapashmajumder: That will also have the added benefit to being able to unit test it.

swift-sdk/AuthManager.swift
```diff
@@ -51,16 +56,43 @@ class AuthManager: IterableInternalAuthManagerProtocol {
...
+    
+    private var expirationRefreshTimer: Timer?
+    
     private var authToken: String?
     
     private var localStorage: LocalStorageProtocol
     
     private let onAuthTokenRequestedCallback: (() -> String?)?
     
+    private func queueAuthTokenExpirationRefresh(_ authToken: String?) {
```
tapashmajumder, line 75: Please add unit test for this.

swift-sdk/AuthManager.swift
```diff
@@ -51,16 +56,43 @@ class AuthManager: IterableInternalAuthManagerProtocol {
...
+        guard let authToken = authToken, let expirationDate = AuthManager.decodeExpirationDateFromAuthToken(authToken) else {
+            return
+        }
+        
+        let refreshDate = TimeInterval(expirationDate - AuthManager.refreshWindow)
+        
+        if #available(iOS 10.0, *) {
+            expirationRefreshTimer = Timer.scheduledTimer(withTimeInterval: refreshDate, repeats: false) { timer in
```
tapashmajumder, line 83: `scheduledTimer` method expects time interval and not a date.

swift-sdk/AuthManager.swift
```diff
@@ -51,16 +56,43 @@ class AuthManager: IterableInternalAuthManagerProtocol {
...
     private var localStorage: LocalStorageProtocol
     
     private let onAuthTokenRequestedCallback: (() -> String?)?
     
+    private func queueAuthTokenExpirationRefresh(_ authToken: String?) {
+        guard let authToken = authToken, let expirationDate = AuthManager.decodeExpirationDateFromAuthToken(authToken) else {
+            return
+        }
+        
+        let refreshDate = TimeInterval(expirationDate - AuthManager.refreshWindow)
```
tapashmajumder, line 80: I am not sure what `refreshDate` is here. Is this a date or time interval?

swift-sdk/AuthManager.swift
roninopf: I'm going to rename it, `refreshDate` is actually a `TimeInterval`.

swift-sdk/AuthManager.swift
roninopf: (same as above)

tapashmajumder:
What is `expirationDate`?

roninopf:
> What is `expirationDate`?

It's the `exp` field within the payload that determines when the given token will not be valid. It's an optional field.

tapashmajumder:
Since `queueAuthToken..` method is not a static one. It should not be made internal just for unit tests if it doesn't need to be. I have checked in a commit to fix this. The unit test needed to be modified slightly so that the callback is invoked.

For `decodeExpirationDateFromAuthToken` static method is OK to remain internal because static methods are simply utility methods with no state. Meaning that we might has well have had this method in a separate utility class such as JWTHelper.

tapashmajumder (approved):
+1

PR #383, opened 2020-09-18
[MOB-2122] make token refresh window configurable

roninopf:


swift-sdk/IterableConfig.swift
```diff
@@ -120,6 +120,9 @@ public class IterableConfig: NSObject {
     /// How many seconds to wait before showing the next in-app, if there are more than one present
     public var inAppDisplayInterval: Double = 30.0
     
+    /// the number of seconds before expiration of the auth token to get a new auth token
+    public var authTokenRefreshWindow: TimeInterval = 60.0
```
tapashmajumder, line 124: The default value seems too aggressive. Is this right? This means we will have to refresh prior to most API calls.

roninopf: Oh, this isn't before every call, it's (by default) 60 seconds before the auth token expires.

tapashmajumder (approved):
+1

PR #388, opened 2020-09-22
[MOB-2126] jwt utilizing completion blocks

roninopf:


tapashmajumder:
It appears that in all the unit tests config.onAuthTokenRequestedCallback returns immediately. Can you please write a test that does not return the authToken synchronously? That is, we test the situation where auth token is returned by client after sometime.

swift-sdk/IterableConfig.swift
```diff
@@ -60,6 +60,10 @@ import Foundation
     func log(level: LogLevel, message: String)
 }
```
tapashmajumder, line 62: Can you please add a couple of lines of documentation here?

swift-sdk/IterableConfig.swift
```diff
@@ -60,6 +60,10 @@ import Foundation
...
 
+@objc public protocol IterableAuthDelegate: AnyObject {
```
tapashmajumder, line 63: Can you please add a couple of lines of documentation for this protocol.

tapashmajumder (approved):
+1

PR #389, opened 2020-09-22
[MOB-1510] update sample apps

roninopf:


tapashmajumder:
Awesome!
Have you verified that these are working as expected?

roninopf:
> Awesome!
> Have you verified that these are working as expected?

Yep! I pushed 1 more commit to remove SDK schemes from the sample apps.

tapashmajumder (approved):
+1

PR #406, opened 2020-10-14
[MOB-2172] - InApp Animation

tapashmajumder:


swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -154,30 +174,95 @@ class IterableHtmlMessageViewController: UIViewController {
...
             }
     }
     
-    private static func safeAreaInsets(for view: UIView) -> UIEdgeInsets {
-        if #available(iOS 11, *) {
-            return view.safeAreaInsets
-        } else {
-            return .zero
+    private func animateWhileEntering(position: ViewPosition) {
+        Self.animate(duration: parameters.animationDuration) { [weak self] in
```
roninopf, line 213: Huh. I've never seen `Self` before. Is it just the same thing as `self`? Seems like it.

roninopf (approved):

swift-sdk/Internal/IterableHtmlMessageViewController.swift
tapashmajumder: `Self` refers to class methods and vars, whereas `self` is for instance.

roninopf: OH. That makes way more sense!

PR #414, opened 2020-10-23
[MOB-2257] only retrieve inbox messages not all in-apps

roninopf:


vbabenkoru (approved):
Wow, great catch @roninopf!

PR #419, opened 2020-10-29
[MOB-2280] exclude plists from targets

roninopf:


roninopf:
Merging since it has been independently verified by the original reporter!

PR #422, opened 2020-11-16
[MOB-2308] xcframework generator scripts

roninopf:


tapashmajumder:
Is this solution working? Have you tested whether the binaries work in simulator as well as devices?

roninopf:
> Is this solution working? Have you tested whether the binaries work in simulator as well as devices?

Tested on all simulator models (e.g. different iPad/iPhone generations), device (when set to optional target membership), archiving, building and running!

tapashmajumder (approved):
+1

PR #430, opened 2020-12-11
[MOB-2365] account for more JWT cases

roninopf:
https://iterable.atlassian.net/browse/MOB-2365
https://iterable.atlassian.net/browse/MOB-2426

### DIFF
--- a/swift-sdk/Internal/AuthManager.swift
+++ b/swift-sdk/Internal/AuthManager.swift
@@ -43,6 +43,8 @@ class AuthManager: IterableInternalAuthManagerProtocol {
     
     // @objc attribute only needed for the pre-iOS 10 Timer constructor in queueAuthTokenExpirationRefresh
     @objc func requestNewAuthToken(hasFailedPriorAuth: Bool = false, onSuccess: AuthTokenRetrievalHandler? = nil) {
+        ITBInfo()
+        
         guard !pendingAuth else {
             return
         }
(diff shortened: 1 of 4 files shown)


tapashmajumder:
Can you explain briefly what changes you have made in this PR? 
Also, do we need any tests to verify that the new JWT cases are working.

roninopf:
> Can you explain briefly what changes you have made in this PR?
> Also, do we need any tests to verify that the new JWT cases are working.

2 changes:
- always make the auth token request if logged in, regardless of "previous" user (before, it wouldn't do so if JWT got turned on later)
- tied to the above change, don't go into the auth token flow if there's no actual user (e.g. logged out)

tapashmajumder:
This looks much better. Thanks for adding the tests.

So, in order to request new auth token, we have to set the email again. This seems semantically quite odd and confusing. 
Why can't we add this method `IterableAPI.requestNewToken()`?

roninopf:
> This looks much better. Thanks for adding the tests.
> 
> So, in order to request new auth token, we have to set the email again. This seems semantically quite odd and confusing.
> Why can't we add this method `IterableAPI.requestNewToken()`?

Oh, it's for the case where the app starts up again and also needs to request a new token; the bug was that if a user was logged into an SDK version WITHOUT JWT, they would never initiate the flow to get a token.

tapashmajumder:
Oh I see. So is a new JWT token will be requested every time the app is initialized for the same user, right? Is that the correct behavior?

roninopf:
> Oh I see. So is a new JWT token will be requested every time the app is initialized for the same user, right? Is that the correct behavior?

Yes.

tapashmajumder (approved):
+1

PR #439, opened 2021-01-07
[MOB-2491] refactor

roninopf:
- moving test files into category folders
- removing (most) author attributions


tapashmajumder:
I am not sure about directory names with spaces in them. Makes it harder to write scripts etc. Consider renaming "Deep Linking Tests" -> "deep-linking-tests". This convention is consistent with out other repositories as well.

roninopf:
> I am not sure about directory names with spaces in them. Makes it harder to write scripts etc. Consider renaming "Deep Linking Tests" -> "deep-linking-tests". This convention is consistent with out other repositories as well.

Fixed

tapashmajumder (approved):
+1

PR #442, opened 2021-01-20
[MOB-2419] in-app priority

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-2419](https://iterable.atlassian.net/browse/MOB-2419)

## ✏️ Description
- adds in-app priority to consideration when showing in-apps


swift-sdk/Constants.swift
```diff
@@ -55,6 +55,15 @@ public enum Const {
...
         static let attributionInfoExpiration = 24
     }
+    
+    enum PriorityLevel {
+        static let critical = 100.0
+        static let high = 200.0
+        static let medium = 300.0
+        static let low = 400.0
+        
+        static let unassigned = 300.5
```
tapashmajumder, line 65: Why is this 300.5 and not `medium`?

tests/common/CommonMocks.swift
```diff
@@ -450,7 +450,7 @@ class MockNotificationCenter: NotificationCenterProtocol {
     private var observers = [Observer]()
 }
 
-class MockInAppPesister: InAppPersistenceProtocol {
```
tapashmajumder, line 453: 😄

swift-sdk/IterableMessaging.swift
```diff
@@ -205,7 +208,8 @@ public extension Notification.Name {
          saveToInbox: Bool = false,
          inboxMetadata: IterableInboxMetadata? = nil,
          customPayload: [AnyHashable: Any]? = nil,
-         read: Bool = false) {
+         read: Bool = false,
+         priorityLevel: Double? = nil) {
```
tapashmajumder, line 212: Since we have a default priority level, I think we should set it to `PriorityLevel.default`. The logic will be much easier if we don't have to deal with `nil`.

swift-sdk/Internal/InAppManager+Functions.swift
```diff
@@ -69,7 +69,10 @@ struct MessagesProcessor {
     }
     
     private func getFirstProcessableTriggeredMessage() -> IterableInAppMessage? {
-        messagesMap.values.filter(MessagesProcessor.isProcessableTriggeredMessage).first
+        messagesMap.values
+            .filter(MessagesProcessor.isProcessableTriggeredMessage)
+            .sorted { ($0.priorityLevel ?? Const.PriorityLevel.unassigned) < ($1.priorityLevel ?? Const.PriorityLevel.unassigned) }
```
tapashmajumder, line 74: This logic will become easier if `priorityLevel` is not `nil`. See comments below.

swift-sdk/IterableMessaging.swift
```diff
@@ -194,6 +194,9 @@ public extension Notification.Name {
         saveToInbox && trigger.type == .never
     }
     
+    /// the urgency level of this message (nil will be treated as `unassigned` when displaying this message)
+    public var priorityLevel: Double?
```
tapashmajumder, line 198: Consider changing this to not be nullable.

tests/common/CommonMocks.swift
roninopf: 😉

swift-sdk/Constants.swift
roninopf: I realized I forgot to finalize the values of the levels and how to utilize them (e.g. `isCritical()`).

`300.5` is the true middle value since it also is right in the middle of the different possible campaign values.

tapashmajumder (approved):
+1

PR #457, opened 2021-03-04
[MOB-2206] - Remove fatalerror

tapashmajumder:
## 🔹 JIRA Ticket(s) if any

* [MOB-2206](https://iterable.atlassian.net/browse/MOB-2206)

## ✏️ Description

> Removed fatalError from places where we could return default or an error.

### DIFF
--- a/swift-sdk/Internal/IterableCoreDataPersistence.swift
+++ b/swift-sdk/Internal/IterableCoreDataPersistence.swift
@@ -47,7 +47,7 @@ class PersistentContainer: NSPersistentContainer {
         let container = PersistentContainer(name: PersistenceConst.dataModelFileName, managedObjectModel: managedObjectModel)
         container.loadPersistentStores { desc, error in
             if let error = error {
-                fatalError("Unresolved error \(error)")
+                ITBError("Unresolved error when creating PersistentContainer: \(error)")
             }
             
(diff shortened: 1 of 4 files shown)


swift-sdk/Internal/IterableCoreDataPersistence.swift
```diff
@@ -47,7 +47,7 @@ class PersistentContainer: NSPersistentContainer {
         let container = PersistentContainer(name: PersistenceConst.dataModelFileName, managedObjectModel: managedObjectModel)
         container.loadPersistentStores { desc, error in
             if let error = error {
-                fatalError("Unresolved error \(error)")
+                ITBError("Unresolved error when creating PersistentContainer: \(error)")
```
roninopf, line 50: Is it okay to just let container continue to be created here? It'll just fall through.

tapashmajumder: This is a callback from a separate thread and we can't really return a nil PersistentContainer here. The container is already created at this point.

We should definitely not crash here with a fatalError so this implementation is better than what was existing. 

Ideally maybe in the future we can do something funky such as making this call asynchronous but that would be a lot more effort and I don't think we need to worry about that.

roninopf (approved):

PR #458, opened 2021-03-09
[MOB-2568] add in-app priority test for sorting on a list

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2568](https://iterable.atlassian.net/browse/MOB-2568)

## ✏️ Description

* add another test for in-app priority

### DIFF
--- a/tests/swift-sdk-swift-tests/in-app-tests/InAppPriorityTests.swift
+++ b/tests/swift-sdk-swift-tests/in-app-tests/InAppPriorityTests.swift
@@ -13,6 +13,51 @@ class InAppPriorityTests: XCTestCase {
                                                          logDelegate: DefaultLogDelegate())
     }
     
+    func testDisplayingPriorityLevelsInOrder() {
+        let condition1 = expectation(description: "not able to match message priority levels")
+        condition1.expectedFulfillmentCount = 4
+        
+        let messages = [
+            getMessageWithPriority("1", Const.PriorityLevel.low),
+            getMessageWithPriority("2", Const.PriorityLevel.high),
+            getMessageWithPriority("3", Const.PriorityLevel.medium),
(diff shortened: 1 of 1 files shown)


tests/swift-sdk-swift-tests/in-app-tests/InAppPriorityTests.swift
```diff
@@ -13,6 +13,51 @@ class InAppPriorityTests: XCTestCase {
...
     
+    func testDisplayingPriorityLevelsInOrder() {
+        let condition1 = expectation(description: "not able to match message priority levels")
+        condition1.expectedFulfillmentCount = 4
+        
+        let messages = [
+            getMessageWithPriority("1", Const.PriorityLevel.low),
+            getMessageWithPriority("2", Const.PriorityLevel.high),
+            getMessageWithPriority("3", Const.PriorityLevel.medium),
+            getMessageWithPriority("4", Const.PriorityLevel.critical)
```
tapashmajumder, line 24: I forget why are priority levels stored in `Const`. I don't see `PriorityLevel.medium`, `PriorityLevel.critical` being used anywhere.

roninopf: I wasn't sure where else to put them (maybe in a dedicated `InAppPriority` Swift file?) but they're there if a developer wants to make comparisons (e.g. show this in-app if `Const.PriorityLevel.critical`)

tapashmajumder: I think these consts have internal visibility. And that is fine. I think all our Consts should be for internal consumption.

I think we should create a separate ticket for exposing and using PriorityLevels, if that is needed.

tapashmajumder (approved):
+1

PR #460, opened 2021-03-11
[MOB-1604] revise service extension to be reactive

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-1604](https://iterable.atlassian.net/browse/MOB-1604)

## ✏️ Description

* revises the notification service extension to not faultily trigger
* refactored to avoid giant singular functions
* added/integrated last attempt and/or failure cases


roninopf:
Sending back to WIP!

notification-extension/ITBNotificationServiceExtension.swift
```diff
@@ -16,129 +16,138 @@ import UserNotifications
...
-            return messageId
-        } else {
-            return content.categoryIdentifier
         }
     }
     
-    private func createNotificationActionButton(buttonDictionary: [AnyHashable: Any]) -> UNNotificationAction? {
-        guard let identifier = buttonDictionary[JsonKey.ActionButton.identifier] as? String else {
-            return nil
+    private func lastAttemptCalled() {
```
tapashmajumder, line 78: It is not very clear from the method name what it does. Is this method something that needs to be always called at the end in case of failure?

notification-extension/ITBNotificationServiceExtension.swift
```diff
@@ -16,129 +16,138 @@ import UserNotifications
...
-                
-                if let attachment = attachment, let bestAttemptContent = strongSelf.bestAttemptContent, let contentHandler = strongSelf.contentHandler {
-                    bestAttemptContent.attachments.append(attachment)
-                    contentHandler(bestAttemptContent)
-                }
-            } else { /* TODO: handle download error */ }
-        })
+    // MARK: - Private
+    
+    private func retrieveAttachment(itblDictionary: [AnyHashable: Any], onFailureHandler: @escaping () -> Void) {
```
tapashmajumder, line 37: I like the name `failureHandler` instead of `onFailureHandler` if it is a closure. I am not sure that we need to pass along a closure? Can't we just call the `lastAttemptCalled` method? Also, maybe also rename `lastAttemptCalled` to `onFailure` or `onCompletion`.

notification-extension/ITBNotificationServiceExtension.swift
roninopf: That's correct; I'll figure out a better name, since it IS referencing the `bestAttemptContent` attribute in the extension.

notification-extension/ITBNotificationServiceExtension.swift
roninopf: These are good suggestions; I wasn't sure if the handler needed to have that flexibility of passing along a closure, so I kept it in. I think you're right that there's no need to do that.

tapashmajumder:
This looks good to me. But please wait for @vbabenkoru to approve before merging.
Also, hopefully you have tested the following scenarios.

1. Multiple buttons including `delete` and `text` type.
2. Rich media url.

tapashmajumder:
@vbabenkoru since you did the initial implementation please take a second look. Thanks!

vbabenkoru (commented):
I haven't tested it but it seems good.
One comment though: make sure it works when buttons are set, but there is no image. There have been reports that sometimes it doesn't render buttons when there is no image, or if there is an image and we've already shown it. My guess is that is because when we set the category (buttons) in this code and return immediately, iOS sometimes doesn't propagate this information and ends up not rendering the buttons. If there is a new image, we have to download it, and that adds some delay that lets the category to get registered by the time we call the content handler. If there is no image or it is cached, we'll return too soon and possibly fail to render the buttons. It'd be good to have some kind of slight delay as a protection against this race condition (either this PR or separately).

roninopf:
Moving this to [WIP] since I ran into a discrepancy between what the current unit tests check for and manual testing cases. It needs a little more investigation to determine the EXACT correction to make.

PR #464, opened 2021-03-30
[MOB-2769] version 6.2.21

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2769](https://iterable.atlassian.net/browse/MOB-2769)

## ✏️ Description

* version 6.2.21 with read state and no messages view


Ayyanchira (commented):
As there are other fixes as well in this stable release, wondering if we should add those fixes in changelog too?

Ayyanchira (approved):

PR #467, opened 2021-04-08
[MOB-2747] fix auth flow with setting user

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2747](https://iterable.atlassian.net/browse/MOB-2747)

## ✏️ Description

* remove travis CI config since it's not used anymore
* fixed auth tests to proper test cases
* fixed auth such that `setEmail` and `setUserId` will do post-login sequencing in proper order (request auth token with completion calling the rest of the login sequence), and removed temporary workaround (can also look at android SDK's implementation for reference)


tapashmajumder:
@roninopf  Please see why tests are failing. Rerun if the tests are flaky.

roninopf:
> @roninopf Please see why tests are failing. Rerun if the tests are flaky.

Thanks for the heads up, it had totally missed me

roninopf:
Verified that changes are working as intended by logging in as brand new users (with corresponding tokens set on the delegate), and checking both the logs and in test project that those users are successfully created and registered.

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -471,6 +451,27 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
...
+        
+        if isEitherUserIdOrEmailSet() && config.authDelegate != nil {
+            requestNewAuthToken()
+        } else {
+            loginNewUser()
+        }
+    }
+    
+    private func requestNewAuthToken() {
+        authManager.requestNewAuthToken(hasFailedPriorAuth: false, onSuccess: { [weak self] authToken in
```
tapashmajumder, line 470: nit: replace `authToken` in callback with `_`

tests/swift-sdk-swift-tests/AuthTests.swift
```diff
@@ -702,8 +702,9 @@ class AuthTests: XCTestCase {
         wait(for: [condition2], timeout: 3)
     }
     
-    func testAuthTokenRequestingForAlreadyExistingEmail() {
-        let condition1 = expectation(description: "auth handler didn't get called")
+    func testAuthTokenNotRequestingForAlreadyExistingEmail() {
+        let condition1 = expectation(description: "auth handler got called when it should not")
+        condition1.isInverted = true
```
tapashmajumder, line 707: We should only wait for a limited time since this is inverted expectation. Use `testExpectationTimeoutForInverted`.

tests/swift-sdk-swift-tests/AuthTests.swift
```diff
@@ -573,7 +573,7 @@ class AuthTests: XCTestCase {
         condition1.expectedFulfillmentCount = 2
         
         let authDelegate = createAuthDelegate({ completion in
-            completion(nil)
+            completion(AuthTests.authToken)
```
tapashmajumder, line 576: I think we should check for expectations here? Otherwise `authDelegate` in this test is basically not being tested at all. We might as well not set it.

tests/swift-sdk-swift-tests/AuthTests.swift
```diff
@@ -179,10 +177,11 @@ class AuthTests: XCTestCase {
         let updatedEmail = "second@example.com"
         let updatedToken = "jay"
         
-        let authDelegate = createAuthDelegate({ completion in
-            if internalAPI?.email == originalEmail { completion(originalToken) }
-            if internalAPI?.email == updatedEmail { completion(updatedToken) }
-        })
+        let authDelegate = DefaultAuthDelegate2 {
```
tapashmajumder, line 180: This is causing a compilation failure.

tapashmajumder (approved):
+1

swift-sdk/Internal/AuthManager.swift
```diff
@@ -101,11 +101,11 @@ class AuthManager: IterableInternalAuthManagerProtocol {
         
         storeAuthToken()
         
+        queueAuthTokenExpirationRefresh(authToken)
+        
         if authToken != nil {
             onSuccess?(authToken)
         }
-        
-        queueAuthTokenExpirationRefresh(authToken)
```
vbabenkoru, line 108: What was the reason for moving this call?

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -471,6 +451,27 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         try? requestHandler.handleLogout()
     }
     
+    private func storeIdentifierData() {
+        localStorage.email = _email
+        localStorage.userId = _userId
+    }
+    
+    private func postIdentifierActions() {
```
vbabenkoru, line 459: I think it's not obvious what this method does from its name without looking at its code... At first I thought "it seems to.. post identifier actions? what does that mean?".

swift-sdk/Internal/AuthManager.swift
roninopf: Semantically there should be no difference but it was to follow the same flow that Android does [here](https://github.com/Iterable/iterable-android-sdk/blob/f698238f800c49b97cd40d892807eca031e5e8a3/iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java#L49).

swift-sdk/Internal/InternalIterableAPI.swift
roninopf: You're right. It kind of sounds like a line from a sci-fi movie. I've renamed it to `postLoginActions`.

vbabenkoru: Oh, I also meant that since most method names are usually verbs, and `post` is an ambiguous word, I start thinking this thing posts login actions... somewhere.

roninopf: Ohhhh! Okay, changed to `onLogin` like on Android. That's clearer.

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -135,8 +122,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         
         storeIdentifierData()
         
-        requestNewAuthToken()
-        loginNewUser()
+        onLogin()
```
tapashmajumder, line 125: So now we have `onLogin` which calls `loginNewUser`. I think it is a little confusing because `onLogin` should be called **after** login is completed. 

I think `loginNewUser` should be called `onLogin` and `onLogin` should be called `loginNewUser`.

Or maybe `loginNewUser` should be renamed to updateBackendOnLogin.

roninopf: Hmm, yeah, I agree. I'll find a new name for `loginNewUser`, I think that would be the best pick for a name revision.

PR #484, opened 2021-06-24
[MOB-2901] add ability to have the inbox badge be numberless

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-2901](https://iterable.atlassian.net/browse/MOB-2901)

## ✏️ Description

* adds an option in the inbox VC to say whether there's a number in the badge, when there are unread messages

### DIFF
--- a/swift-sdk/IterableInboxViewController.swift
+++ b/swift-sdk/IterableInboxViewController.swift
@@ -119,7 +119,10 @@ open class IterableInboxViewController: UITableViewController {
 
     /// Use this to set the message to show when there are no message in the inbox.
     @IBInspectable public var noMessagesBody: String? = nil
-
+    
+    /// If `true`, the inbox badge will show a number when there are any unread messages in the inbox.
+    /// If `false` it will simply show an indicator if there are any unread messages in the inbox.
+    @IBInspectable public var showCountInUnreadBadge: Bool = true
     
     /// when in popup mode, specify here if you'd like to change the presentation style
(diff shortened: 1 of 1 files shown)


Ayyanchira (approved):

tapashmajumder:
I think this change will not solve the problem in the Jira ticket. 

If I understand correctly, this change will either display a dot or a number in the tab bar icon if there are unread messages.

What the customer seems to want is to be able to completely not show any indicator in certain cases.

swift-sdk/IterableInboxViewController.swift
```diff
@@ -433,8 +436,14 @@ extension IterableInboxViewController: InboxViewControllerViewModelView {
...
         let unreadCount = viewModel.unreadCount
-        let badgeValue = unreadCount == 0 ? nil : "\(unreadCount)"
-        navigationController?.tabBarItem?.badgeValue = badgeValue
+        let badgeUnreadCount = unreadCount == 0 ? nil : "\(unreadCount)"
+        let badgeUnreadIndicator = unreadCount == 0 ? nil : ""
+        
+        if showCountInUnreadBadge {
+            navigationController?.tabBarItem?.badgeValue = badgeUnreadCount
+        } else {
+            navigationController?.tabBarItem?.badgeValue = badgeUnreadIndicator
```
tapashmajumder, line 445: This should be nil. The customer does not want to show either a "dot or a badge". This will probably show a dot.

tapashmajumder (changes requested):
Please take a look at my comments.

PR #487, opened 2021-07-21
[MOB-725] updateCart (part 2)

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-725](https://iterable.atlassian.net/browse/MOB-725)

## ✏️ Description

* Adds the [updateCart API](https://api.iterable.com/api/docs#commerce_updateCart) into our SDK
* The part 2 of the other PR which fixes and adds test case related to the part 1 PR. It also has a bunch of quick refactoring that was convenient while I was working on these files.

### DIFF
--- a/swift-sdk/Internal/RequestCreator.swift
+++ b/swift-sdk/Internal/RequestCreator.swift
@@ -16,16 +16,20 @@ struct RequestCreator {
     // MARK: - API REQUEST CALLS
     
     func createUpdateEmailRequest(newEmail: String) -> Result<IterableRequest, IterableError> {
-        var body: [String: Any] = [JsonKey.newEmail: newEmail]
+        if case .none = auth.emailOrUserId {
+            ITBError(authFailedMessage)
+            return .failure(IterableError.general(description: authFailedMessage))
+        }
+        
+        var body = [String: Any]()
         
         if let email = auth.email {
             body[JsonKey.currentEmail] = email
         } else if let userId = auth.userId {
             body[JsonKey.currentUserId] = userId
-        } else {
-            ITBError("Both email and userId are nil")
-            return .failure(IterableError.general(description: "Both email and userId are nil"))
         }
+        
+        body[JsonKey.newEmail] = newEmail
         
         return .success(.post(createPostRequest(path: Const.Path.updateEmail, body: body)))
     }
@@ -33,10 +37,10 @@ struct RequestCreator {
     func createRegisterTokenRequest(registerTokenInfo: RegisterTokenInfo,
                                     notificationsEnabled: Bool) -> Result<IterableRequest, IterableError> {
         if case .none = auth.emailOrUserId {
-            ITBError("Both email and userId are nil")
-            return .failure(IterableError.general(description: "Both email and userId are nil"))
+            ITBError(authFailedMessage)
+            return .failure(IterableError.general(description: authFailedMessage))
         }
-
+        
         let dataFields = DataFieldsHelper.createDataFields(sdkVersion: registerTokenInfo.sdkVersion,
                                                            deviceId: registerTokenInfo.deviceId,
                                                            device: UIDevice.current,
@@ -47,7 +51,7 @@ struct RequestCreator {
         let deviceDictionary: [String: Any] = [
             JsonKey.token: registerTokenInfo.hexToken,
             JsonKey.platform: RequestCreator.pushServicePlatformToString(registerTokenInfo.pushServicePlatform,
-                                                                                 apnsType: registerTokenInfo.apnsType),
+                                                                         apnsType: registerTokenInfo.apnsType),
             JsonKey.applicationName: registerTokenInfo.appName,
             JsonKey.dataFields: dataFields,
         ]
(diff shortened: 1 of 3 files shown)


swift-sdk/Internal/RequestCreator.swift
```diff
@@ -16,27 +16,31 @@ struct RequestCreator {
...
+        
         return .success(.post(createPostRequest(path: Const.Path.updateEmail, body: body)))
     }
     
     func createRegisterTokenRequest(registerTokenInfo: RegisterTokenInfo,
                                     notificationsEnabled: Bool) -> Result<IterableRequest, IterableError> {
         if case .none = auth.emailOrUserId {
-            ITBError("Both email and userId are nil")
-            return .failure(IterableError.general(description: "Both email and userId are nil"))
+            ITBError(authFailedMessage)
```
tapashmajumder, line 40: 'authFailedMessage' is a generic name for variable. This is a very specific one so we should name it differently. 'authMissing', 'userCredentialMissing'  or of that kind.

swift-sdk/Internal/RequestCreator.swift
```diff
@@ -425,12 +432,14 @@ struct RequestCreator {
...
             args[JsonKey.InApp.packageName] = packageName
         }
-
+        
         return .success(.get(createGetRequest(forPath: Const.Path.getRemoteConfiguration, withArgs: args as! [String: String])))
     }
     
     // MARK: - PRIVATE
     
+    private let authFailedMessage = "Both email and userId are nil"
```
tapashmajumder, line 441: I think static would be better since this has no instance specific information. So basically
`private static let authMissingMessage` and in other places `Self.authMissingMessage`. (notice capital S).

tests/unit-tests/IterableAPITests.swift
```diff
@@ -405,8 +405,7 @@ class IterableAPITests: XCTestCase {
             }
         }
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 408: Expectation does not seem to be inverted here.

tests/unit-tests/IterableAPITests.swift
```diff
@@ -424,8 +423,7 @@ class IterableAPITests: XCTestCase {
             expectation.fulfill()
         }
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 426: +1

tests/unit-tests/IterableAPITests.swift
```diff
@@ -463,8 +461,7 @@ class IterableAPITests: XCTestCase {
             }
         }
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 464: This is not inverted expectation, the comment is wrong.

tests/unit-tests/IterableAPITests.swift
```diff
@@ -499,8 +496,7 @@ class IterableAPITests: XCTestCase {
             internalAPI.disableDeviceForCurrentUser()
         }
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 499: Not inverted

tests/unit-tests/IterableAPITests.swift
```diff
@@ -539,8 +535,7 @@ class IterableAPITests: XCTestCase {
         
         internalAPI.register(token: token)
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 538: Check if it is actually inverted. Comment may be wrong.

tests/unit-tests/IterableAPITests.swift
```diff
@@ -577,8 +572,39 @@ class IterableAPITests: XCTestCase {
         }
         internalAPI.register(token: token)
         
-        // only wait for small time, supposed to error out
-        wait(for: [expectation], timeout: testExpectationTimeout)
+        wait(for: [expectation], timeout: testExpectationTimeoutForInverted)
```
tapashmajumder, line 575: Check if it is actually inverted. Comment may be wrong.

tests/unit-tests/IterableAPITests.swift
```diff
@@ -577,8 +572,39 @@ class IterableAPITests: XCTestCase {
...
+                return
+            }
+            
+            TestUtils.validate(request: request, requestType: .post, apiEndPoint: Endpoint.api, path: Const.Path.updateCart, queryParams: [])
+            TestUtils.validateMatch(keyPath: KeyPath(string: "\(JsonKey.Commerce.user).\(JsonKey.email)"), value: IterableAPITests.email, inDictionary: body)
+            
+            let itemsElement = body[JsonKey.Commerce.items] as! [[AnyHashable: Any]]
+            XCTAssertEqual(itemsElement.count, items.count)
+            
+            //TODO: create a CommerceItem matcher for use right here, and in trackPurchase tests
```
tapashmajumder, line 600:
```suggestion
            // TODO: create a CommerceItem matcher for use right here, and in trackPurchase tests
```

tapashmajumder, line 600: Do we need to add CommerceItem matcher for this PR?

tapashmajumder:
It appears that in many places in IterableAPITests the comments are wrong but the code is OK. The expectations are indeed not inverted. Please check everywhere that you have changed to make sure that the timeout is for inverted only for inverted expectations.

tests/unit-tests/IterableAPITests.swift
roninopf: Functionally, we don't (the test still checks the correctness of the items), but I just saw a refactoring possibility here, in the future.

PR #497, opened 2021-08-12
[MOB-3251] add header_dir to podspec

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-3251](https://iterable.atlassian.net/browse/MOB-3251)

## ✏️ Description

* add `header_dir` to the SDK Podspec so that `use_frameworks!` doesn't need to be in the Podfile anymore


tapashmajumder:
Can you please verify that all our sample apps that use cocoapods are working fine after this change?

roninopf:
> Can you please verify that all our sample apps that use cocoapods are working fine after this change?

We don't have any sample apps using Cocoapods, but I did test this with a brand new iOS project beforehand.

tapashmajumder:
Can you please add more details about the solution? Why is this change needed and what does it do?

tapashmajumder (approved):
+1

PR #512, opened 2021-11-01
[MOB-3596] - Keychain to store JWT token

tapashmajumder:
## 🔹 Jira Ticket(s)

* [MOB-3596](https://iterable.atlassian.net/browse/MOB-XXXX)

## ✏️ Description

> We need to store JWT auth token in keychain instead of UserDefaults. More details are here
https://iterable.slab.com/posts/cleartext-jwt-storage-remediation-wnj6gj9y.

### DIFF
--- a/swift-sdk/Internal/LocalStorage.swift
+++ b/swift-sdk/Internal/LocalStorage.swift
@@ -1,0 +1,96 @@
+//
+//  Copyright © 2018 Iterable. All rights reserved.
+//
+
+import Foundation
+
+struct LocalStorage: LocalStorageProtocol {
+    init(userDefaults: UserDefaults = UserDefaults.standard) {
+        iterableUserDefaults = IterableUserDefaults(userDefaults: userDefaults)
+    }
+    
+    var userId: String? {
+        get {
+            iterableUserDefaults.userId
+        } set {
+            iterableUserDefaults.userId = newValue
+        }
+    }
+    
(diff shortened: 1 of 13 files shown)


swift-sdk/Internal/LocalStorage.swift
```diff
@@ -0,0 +1,108 @@
...
+        iterableUserDefaults.getPayload(currentDate: currentDate)
+    }
+    
+    func save(payload: [AnyHashable: Any]?, withExpiration expiration: Date?) {
+        iterableUserDefaults.save(payload: payload, withExpiration: expiration)
+    }
+    
+    func upgrade() {
+        ITBInfo()
+        if let userDefaultAuthToken = iterableUserDefaults.authToken, keychain.authToken == nil {
```
roninopf, line 97: Not required for approval, but should this be separated into its own function, or be delineated as a full block of code handling the JWT transfer?

roninopf (approved):

swift-sdk/Internal/LocalStorage.swift
tapashmajumder: @roninopf  I like this idea. Do you have any suggestion regarding what we should call this function?

roninopf: Hmm, maybe like `moveJwtFromUserDefaultsToKeychain`?

PR #520, opened 2021-12-06
[MOB-3725] - Allow opening of url from universal links also.

tapashmajumder:
## 🔹 Jira Ticket(s)

* [MOB-3725](https://iterable.atlassian.net/browse/MOB-3725)

## ✏️ Description

> Allow opening urls from universal links also.

### DIFF
--- a/swift-sdk/Internal/IterableActionRunner.swift
+++ b/swift-sdk/Internal/IterableActionRunner.swift
@@ -77,8 +77,7 @@ struct IterableActionRunner {
     private static func shouldOpenUrl(url: URL,
                                       from source: IterableActionSource,
                                       withAllowedProtocols allowedProtocols: [String]) -> Bool {
-        if source == .push || source == .inApp,
-           let scheme = url.scheme,
+        if let scheme = url.scheme,
            scheme == "https" || allowedProtocols.contains(scheme) {
             return true
         } else {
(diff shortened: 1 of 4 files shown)


swift-sdk/Internal/IterableActionRunner.swift
```diff
@@ -77,8 +77,7 @@ struct IterableActionRunner {
     private static func shouldOpenUrl(url: URL,
                                       from source: IterableActionSource,
                                       withAllowedProtocols allowedProtocols: [String]) -> Bool {
-        if source == .push || source == .inApp,
-           let scheme = url.scheme,
+        if let scheme = url.scheme,
```
roninopf, line 80: Not required for approval, but I think it would look cleaner using `guard`

roninopf (approved):

swift-sdk/Internal/IterableActionRunner.swift
tapashmajumder: I think normally `guard` statements are for early exits or short circuiting code, so that  exceptions don't up the logic of the method   body. Here since the is no method body as such I  think if is better.

tapashmajumder: In other words, `guard` is primarily to make sure that certain conditions are met before executing deeper into the method.

PR #539, opened 2022-03-29
XCFramework Fixes and Mac Catalyst Support

teamplayon:
## ✏️ Description

- Fixes a build issue when using the create-xcframework-zips.sh script due to i386 being enabled but some newer features are not supported and thus need excluded from build just like they are for arm.
- Fixes a regression caused by the implementation of pods/spm support that breaks loading of IterableDataModel.momd when used from a framework or xcframework build.
- Adds support for Mac Catalyst and includes the additional framework in the xcframework build.
- Fixes the create-xcframework-zips.sh error handling such that any failure halts the build rather than blindly continuing and possibly producing erroneous packages.


tapashmajumder:
Hi. All your suggested changes are available with release 6.4.2. Closing this. Please open a Jira Ticket if anything is missing.

PR #540, opened 2022-03-30
[MOB-4239] fix update email with token

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-4239](https://iterable.atlassian.net/browse/MOB-4239)

## ✏️ Description

* fixes functionality to allow setting the JWT for the new email in `updateEmail`


swift-sdk/Internal/AuthManager.swift
```diff
@@ -53,6 +53,12 @@ class AuthManager: IterableAuthManagerProtocol {
         }
     }
     
+    func setNewTokenAfterUpdateEmail(_ newToken: String) {
```
tapashmajumder, line 56: I think we should change the method name to reflect more what it actually does. It appears that this method does not need to be aware that it is called after update email.

tapashmajumder, line 56:
```suggestion
    func setNewToken(_ newToken: String) {
```

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -475,14 +486,23 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
...
         ITBInfo()
         
-        if isEitherUserIdOrEmailSet() && config.authDelegate != nil {
+        if let authToken = authToken {
+            self.authManager.setNewTokenAfterUpdateEmail(authToken)
+        } else if isEitherUserIdOrEmailSet() && config.authDelegate != nil {
             requestNewAuthToken()
-        } else {
-            completeUserLogin()
+            return
```
tapashmajumder, line 496: We should try to exit only in the beginning of method. Exiting in the middle is a code smell.

tapashmajumder: You can move `completeUserLogin` after line 493 and then add an else clause with `completeUserLogin` added there.

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -475,14 +486,23 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         localStorage.userId = _userId
     }
     
-    private func onLogin() {
+    private func onLogin(_ authToken: String? = nil) {
         ITBInfo()
         
-        if isEitherUserIdOrEmailSet() && config.authDelegate != nil {
+        if let authToken = authToken {
```
tapashmajumder, line 492:
```suggestion
        if let authToken = authToken {
           ... setNwAuthToken
           completeUserLogin()
         else if ..... {
            requestNewAuthToken()
         } else {
           completeUserLogin()
         }
```

tests/unit-tests/AuthTests.swift
```diff
@@ -211,6 +211,50 @@ class AuthTests: XCTestCase {
         wait(for: [condition1], timeout: testExpectationTimeout)
     }
     
+    func testUpdateEmailWithTokenParam() {
```
tapashmajumder, line 214: 👏

tapashmajumder:
Logic looks good. I have added some suggestions.

tests/unit-tests/AuthTests.swift
roninopf: :)

tapashmajumder (approved):
+1 Conditional approval based on doing actual testing. Let me know if you need any help

PR #549, opened 2022-04-07
[MOB-4287] Removes the extra api-key argument from the request creator.

davidtruong:
## 🔹 Jira Ticket(s)

* [MOB-4287](https://iterable.atlassian.net/browse/MOB-4287)

## ✏️ Description

> Removes the api-key argument from the url and only passes it in the header of the request to ensure security.

### DIFF
--- a/swift-sdk/Internal/RequestCreator.swift
+++ b/swift-sdk/Internal/RequestCreator.swift
@@ -449,7 +449,7 @@ struct RequestCreator {
     
     private func createPostRequest(path: String, body: [AnyHashable: Any]? = nil) -> PostRequest {
         PostRequest(path: path,
-                    args: [JsonKey.Header.apiKey: apiKey],
+                    args: nil,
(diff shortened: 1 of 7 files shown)


swift-sdk/Internal/RequestCreator.swift
```diff
@@ -449,7 +449,7 @@ struct RequestCreator {
     
     private func createPostRequest(path: String, body: [AnyHashable: Any]? = nil) -> PostRequest {
         PostRequest(path: path,
-                    args: [JsonKey.Header.apiKey: apiKey],
+                    args: nil,
```
tapashmajumder, line 452: I don't think we need `apiKey` for this class anymore. This seems to be the only place apiKey is being used. We should probably remove it.

tapashmajumder (approved):

davidtruong:
merging in changes, tested by Tapash

PR #565, opened 2022-06-21
[MOB-4436] encrypt auth data

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-4436](https://iterable.atlassian.net/browse/MOB-4436)

## ✏️ Description

* adds `email` and `userId` into our keychain

### DIFF
--- a/swift-sdk/Internal/LocalStorage.swift
+++ b/swift-sdk/Internal/LocalStorage.swift
@@ -13,17 +13,17 @@ struct LocalStorage: LocalStorageProtocol {
     
     var userId: String? {
         get {
-            iterableUserDefaults.userId
+            keychain.userId
         } set {
-            iterableUserDefaults.userId = newValue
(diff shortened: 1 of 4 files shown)


swift-sdk/Internal/LocalStorage.swift
```diff
@@ -94,19 +94,36 @@ struct LocalStorage: LocalStorageProtocol {
```
tapashmajumder, line 94: You need to read and write from keychain here.

tapashmajumder: Maybe adding some tests here to catch problems like this.

roninopf: Added a general test to verify presence in the keychain, and the lack of presence in the user defaults.

tapashmajumder (approved):
+1

PR #574, opened 2022-09-01
[MOB-4786] split out mocks

roninopf:
## 🔹 Jira Ticket(s)

* [MOB-4786](https://iterable.atlassian.net/browse/MOB-4786)

## ✏️ Description

* splits out the various classes/structs in `CommonMocks.swift` into their own files
* this is purely a refactor, so there should be 0 functional changes in the SDK or the tests


tests/common/MockNetworkSession.swift
```diff
@@ -227,3 +227,25 @@ class MockNetworkSession: NetworkSessionProtocol {
         return nil
     }
 }
+
```
evangreer91, line 230: Should this be named NoNetworkNetworkSession to align with the class name?

evangreer91: Oh I see this is added to another file

evangreer91 (approved):
LGTM. Looked through and confirmed each mock had the proper imports.

PR #583, opened 2022-11-03
[MOB-4987] split out register device token call

evangreer91:
## 🔹 Jira Ticket(s)

* [MOB-4987](https://iterable.atlassian.net/browse/MOB-4987)

## ✏️ Description

This pull request includes splitting out the register device token call to take in a hex string as well as a Data object.


roninopf:
Will close for now until we revisit this.

PR #586, opened 2022-12-22
[MOB-5337] - InAppDisplayInterval resolution

Ayyanchira:
## 🔹 Jira Ticket(s)

* [MOB-5337](https://iterable.atlassian.net/browse/MOB-5337)

## ✏️ Description

1. Replace instances of retryInterval to moveToForegroundSyncInterval.
2. Map inAppDisplayInterval to moveToForegroundSyncInterval rather than converting to retryInterval

Context:
The configurable inAppDisplayInteval maps to variable - retryInterval which is used to check if its ready to show inapp after first inapp was displayed. However, the same property is not used when app switches from background to foreground. During this time, a variable - moveToForegroundSyncInterval is used which is redundant and is never changed.

### DIFF
--- a/swift-sdk/Internal/InAppManager.swift
+++ b/swift-sdk/Internal/InAppManager.swift
@@ -40,7 +40,7 @@ class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
          applicationStateProvider: ApplicationStateProviderProtocol,
          notificationCenter: NotificationCenterProtocol,
          dateProvider: DateProviderProtocol,
-         retryInterval: Double) {
+         moveToForegroundSyncInterval: Double) {
         ITBInfo()
         
         self.requestHandler = requestHandler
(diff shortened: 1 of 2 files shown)


swift-sdk/Internal/InAppManager.swift
```diff
@@ -532,7 +531,7 @@ class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
     
     private var syncResult: Pending<Bool, Error>?
     private var lastSyncTime: Date?
-    private let moveToForegroundSyncInterval: Double = 1.0 * 60.0 // don't sync within sixty seconds
+    private var moveToForegroundSyncInterval: Double = 1.0 * 60.0 // don't sync within sixty seconds
```
roninopf, line 534: Was this change needed?

roninopf (approved):

swift-sdk/Internal/InAppManager.swift
Ayyanchira: Yes or the tests were failing when not being able to set a custom inapp interval

PR #620, opened 2023-04-04
[MOB-5869] adds data center configuration

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-5869](https://iterable.atlassian.net/browse/MOB-5869)

## ✏️ Description

- create the IterableDataRegion enum which contains the associated endpoint
- adds dataRegion value to the IterableConfig class 

[MOB-5869]: https://iterable.atlassian.net/browse/MOB-5869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit-tests/IterableConfigTests.swift
```diff
@@ -0,0 +1,30 @@
...
+//  File.swift
+//  
+//
+//  Created by evan.greer@iterable.com on 4/5/23.
+//
+
+import XCTest
+
+final class IterableConfigTests: XCTestCase {
+    func testDefaultConfig() {
```
roninopf, line 11: Instead of testing the default config as a whole (which actually has defined defaults), I would just make a test instantiating a default API (using `initializeForTesting`) and then checking that it points to `IterableDataRegion.US`. And then another with the config pointing to EU, and check that the API keeps the EU value. Or wait to do the test until after the endpoints get integrated.

roninopf: I would remove it for now and add a new one when the endpoint changing code is in.

roninopf (approved):

PR #625, opened 2023-04-13
Added delegate that handles rich push from another provider

christinaschell:
## 🔹 Jira Ticket(s)

* [PRO-7882](https://iterable.atlassian.net/browse/PRO-7882)

## ✏️ Description

> Added the `ITBNotificationServiceExtensionDelegate`for the ability to handle rich push from another while still being able to display rich push through Iterable

[PRO-7882]: https://iterable.atlassian.net/browse/PRO-7882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/notification-extension/ITBNotificationServiceExtension.swift
+++ b/notification-extension/ITBNotificationServiceExtension.swift
@@ -4,14 +4,30 @@
 
 import UserNotifications
 
+public protocol ITBNotificationServiceExtensionDelegate {
+    func notificationServiceDidReceive(_ request: UNNotificationRequest, bestAttemptContent: UNMutableNotificationContent, contentHandler: @escaping (UNNotificationContent) -> Void)
+}
+
 @objc open class ITBNotificationServiceExtension: UNNotificationServiceExtension {
     var contentHandler: ((UNNotificationContent) -> Void)?
     var bestAttemptContent: UNMutableNotificationContent?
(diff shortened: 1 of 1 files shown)


notification-extension/ITBNotificationServiceExtension.swift
```diff
@@ -4,15 +4,31 @@
...
     
     @objc override open func didReceive(_ request: UNNotificationRequest,
                                         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
         self.contentHandler = contentHandler
         bestAttemptContent = (request.content.mutableCopy() as? UNMutableNotificationContent)
         
+        if let itblNotificationDelegate = ITBNotificationServiceExtension.itblNotificationDelegate,
+           let bestAttemptContent = bestAttemptContent {
+            
+            let result = request.content.userInfo.contains { $0.key as! String == JsonKey.Payload.metadata }
```
roninopf, line 24: @christinaschell does it make sense to add a failure condition here for casting the key to a `String`? I'm thinking if the cast isn't successful, we print an error to the console and just `return`. I'm thinking that, in the case this cast doesn't succeed, we don't want it to crash here, specifically.

roninopf (approved):

PR #629, opened 2023-04-26
[MOB-6055] callback on setEmail/setUserId

devcsomnicg:
## 🔹 Jira Ticket(s)

* [MOB-6055](https://iterable.atlassian.net/browse/MOB-6055)

## ✏️ Description

> Add callbacks for setEmail/setUserId, and it returns success/failure based on device token is registered successfully or not. It will always return success if autoPushRegistration is false.

[MOB-6055]: https://iterable.atlassian.net/browse/MOB-6055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


devcsomnicg:
@roninopf  Could you please start review for this ?

roninopf:
@devcsomnicg Will do! Would you like to look over the tests as well?

devcsomnicg:
@roninopf updated code and tests, could you please run the tests again

Ayyanchira:
Will have to resolve the conflict. But looks like its replacing the existing method with one with additional parameter. As like for other SDKs, we should have additional methods for the ones with completionHandlers. This is to avoid breakages for developers who would just update to newer versions of SDK

devcsomnicg:
@Ayyanchira done, added additional methods.

devcsomnicg:
@Ayyanchira Updated this with suggestions on PR. please review

devcsomnicg:
@Ayyanchira Updated this with suggestions on PR. please review

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -111,7 +111,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         _payloadData = data
     }
     
-    func setEmail(_ email: String?, authToken: String? = nil) {
+    func setEmail(_ email: String?, authToken: String? = nil, successHandler: OnSuccessHandler? = nil, failureHandler: OnFailureHandler? = nil) {
```
Ayyanchira, line 114: Had initial thought that we might need additional functions with and without handlers as we do in Android. But looks like having optional parameters makes it modern and lets us have only one signature instead of bloating up the IterableApi class. 👍🏼

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -428,6 +440,9 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
     private var _email: String?
     private var _payloadData: [AnyHashable: Any]?
     private var _userId: String?
+    private var _successCallback: OnSuccessHandler? = nil
```
Ayyanchira, line 443: If this is going to be global but only used for setting the user credentials, we might have to rename it that way. Because, we will be adding success failure callbacks to other api calls / function calls as well and they will have to pass in success and failure callbacks too. using this global variable to store that and replace whats been stored in that can cause inconsistent behaviors

Ayyanchira: Like for example, updateEmail also uses success and failure callback handler to notify the state. But by not having it stored globally, it allows for good separation.

Ayyanchira (commented):
Added some comments. Merging to omni cg branch for internal testing. Looks good so far.

swift-sdk/Internal/InternalIterableAPI.swift
devcsomnicg: @Ayyanchira yes we can rename it so that is identified differently, but it would be class variable because it is used from the whole different method and that is why we need to hold a reference to it

swift-sdk/Internal/InternalIterableAPI.swift
devcsomnicg: @Ayyanchira ok so for this we are good right ?

Ayyanchira:
Test methods are failing @devcsomnicg . It looks like setEmail leads to `AppExtensionHelper.application?.registerForRemoteNotifications()`, but that doesn't lead to registerDevice Flow.

devcsomnicg:
> Test methods are failing @devcsomnicg . It looks like setEmail leads to `AppExtensionHelper.application?.registerForRemoteNotifications()`, but that doesn't lead to registerDevice Flow.

I am not sure about where this code is AppExtensionHelper.application.registerForRemoteNotifications() but, to give you a example when you do registerForRemoteNotifications it calls this function didRegisterForRemoteNotificationsWithDeviceToken here https://github.com/Iterable/swift-sdk/blob/5c033ded01e200acee52c9bd1fca377dbdcbbbce/sample-apps/swift-sample-app/swift-sample-app/AppDelegate.swift#L105.

Ayyanchira:
Added a call to registerDevice() with no success failure callback after setEmail -> here in this [PR](https://github.com/Iterable/swift-sdk/pull/641). This [commit](https://github.com/Iterable/swift-sdk/pull/641/commits/01d7e083448439984360e6401ec83253a6d082b6).
Verified that it eventually triggers the callbacks. 
This means the test assumes that registerDevice will eventually get called. And it checks if the handlers set while setting the credentials still remain active or not.

PR #640, opened 2023-05-31
[MOB-6255] Embedded Impression Session

anlinguist:
## 🔹 Jira Ticket(s)

* [MOB-6255](https://iterable.atlassian.net/browse/MOB-6255)

## ✏️ Description

This PR creates an embedded message session that the app can use to manage impression tracking. The SDK handles the API calls and payload, but the app needs to still handle the impression tracking via SDK methods.

[MOB-6255]: https://iterable.atlassian.net/browse/MOB-6255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/EmbeddedSessionManager.swift
+++ b/swift-sdk/Internal/EmbeddedSessionManager.swift
@@ -9,16 +9,16 @@ import Foundation
 
 public class EmbeddedSessionManager {
     public static let shared = EmbeddedSessionManager()
-    public var currentSession: IterableEmbeddedSession?
-    var currentlyTrackingImpressions: [String: (currentDisplayDuration: TimeInterval, startTime: Date, tracking: Bool)] = [:]
+    public var session: IterableEmbeddedSession?
+    var currentlyTrackingImpressions: [String: (totalDisplayDuration: TimeInterval, startTime: Date, tracking: Bool)] = [:]
 
     public func startSession() {
         let startTime = Date()
-        currentSession = IterableEmbeddedSession(embeddedSessionId: UUID().uuidString, embeddedSessionStart: startTime, impressions: [])
+        session = IterableEmbeddedSession(embeddedSessionId: UUID().uuidString, embeddedSessionStart: startTime, impressions: [])
     }
 
     public func endSession() {
-        guard let currentSession = currentSession else {
+        guard let session = session else {
             ITBError("No current session.")
             return
         }
@@ -26,16 +26,16 @@ public class EmbeddedSessionManager {
         for impressionId in currentlyTrackingImpressions.keys {
             pauseImpression(impressionId: impressionId)
         }
-        currentSession.embeddedSessionEnd = Date()
+        session.embeddedSessionEnd = Date()
         updateDisplayDurations()
-        let _ = IterableAPI.embeddedMessagingManager.track(embeddedSession: currentSession)
+        let _ = IterableAPI.embeddedMessagingManager.track(embeddedSession: session)
     }
     
     public func pauseImpression(impressionId: String) {
         if var trackingImpression = currentlyTrackingImpressions[impressionId], trackingImpression.tracking {
             let currentTime = Date()
             let elapsedTime = currentTime.timeIntervalSince(trackingImpression.startTime)
-            trackingImpression.currentDisplayDuration += elapsedTime
+            trackingImpression.totalDisplayDuration += elapsedTime
             trackingImpression.tracking = false
             currentlyTrackingImpressions[impressionId] = trackingImpression
         }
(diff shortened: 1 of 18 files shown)


swift-sdk/Internal/EmbeddedSessionManager.swift
```diff
@@ -0,0 +1,86 @@
...
+        if let timer = messageTimers[impressionId], timer.isValid {
+            return
+        }
+        if pauseTimestamps.removeValue(forKey: impressionId) != nil {
+            if let index = currentSession?.impressions.firstIndex(where: { $0.messageId == impressionId }) {
+                let newDisplayDuration = currentSession!.impressions[index].displayDuration
+                currentSession?.impressions[index].displayDuration = newDisplayDuration
+            }
+            let timer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true) { [weak self] _ in
+                self?.updateDisplayDurationForImpression(impressionId: impressionId)
```
Ayyanchira, line 55: Hmm so theres a method called every one second for every on going impression? what if resume and pause were actually just flags and when pausing, resuming and ending an impression we just calculate the time difference and add/update and store them?

swift-sdk/Internal/MiscEmbeddedClasses.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  File.swift
+//  
+//
+//  Created by Andrew Nelson on 5/24/23.
+//
+
```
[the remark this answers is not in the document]
anlinguist: I initially modeled ImpressionSession after InboxSession. InboxSession uses a MiscInboxClasses.swift file to do similar logic :)

swift-sdk/Internal/EmbeddedSessionManager.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    public func endSession() {
+        currentSession?.embeddedSessionEnd = Date()
+        for timer in messageTimers.values {
+            timer.invalidate()
+        }
+        messageTimers.removeAll()
+        
+        if let session = currentSession {
+            let _ = IterableAPI.embeddedMessagingManager.track(embeddedSession: session)
```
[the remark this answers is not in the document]
anlinguist: I think it makes sense to start the session when app is opened, add/update impressions as needed, and then end session when the app is backgrounded, expecting that the endSession() call should include track the session. User shouldn't need to access `track(embeddedSession:)`. However, this is slightly at odds with the InboxSession approach. The `InboxViewControllerViewModel` has [an endSession() method](https://github.com/Iterable/swift-sdk/blob/0a1e2fe61e0ba28ad09b7ed6a54d01ddcd189e3f/swift-sdk/Internal/InboxViewControllerViewModel.swift#L250) which first [calls the session manager endSession() method](https://github.com/Iterable/swift-sdk/blob/0a1e2fe61e0ba28ad09b7ed6a54d01ddcd189e3f/swift-sdk/Internal/InboxViewControllerViewModel.swift#L252) and [then tracks it](https://github.com/Iterable/swift-sdk/blob/0a1e2fe61e0ba28ad09b7ed6a54d01ddcd189e3f/swift-sdk/Internal/InboxViewControllerViewModel.swift#L266). So Perhaps we should take this approach instead.

swift-sdk/Internal/EmbeddedSessionManager.swift
anlinguist: That's a great idea. I'm refactoring the class to use another time approach via timestamps and flags

swift-sdk/Internal/EmbeddedSessionManager.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+//  Created by Andrew Nelson on 5/30/23.
+//
+
+import Foundation
+
+public class EmbeddedSessionManager {
+    public static let shared = EmbeddedSessionManager()
+    public var currentSession: IterableEmbeddedSession?
+    var currentlyTrackingImpressions: [String: (currentDisplayDuration: TimeInterval, startTime: Date, tracking: Bool)] = [:]
```
Ayyanchira, line 13: `totalDisplayDuration` instead of `currentDisplayDuration` as we are accumulating the total over time.

swift-sdk/Internal/EmbeddedSessionManager.swift
```diff
@@ -0,0 +1,74 @@
...
+//  
+//
+//  Created by Andrew Nelson on 5/30/23.
+//
+
+import Foundation
+
+public class EmbeddedSessionManager {
+    public static let shared = EmbeddedSessionManager()
+    public var currentSession: IterableEmbeddedSession?
```
Ayyanchira, line 12: `session` instead of `currentSession` as there we have only one session one time and no different states of session?

Ayyanchira (approved):
Looking great Andrew. Refactored really well. Added some renaming nit pick comments.

PR #642, opened 2023-06-07
Fix custom payload decoding

anlinguist:
This PR fixes the custom payload decoding in embedded messages. Should work on most standard JSON structures. Tested with the following payload:
```
{
  "test": null,
  "testObj": [
    {
      "testObjKey1": 1,
      "testObjKey2": "two"
    },
    {
      "testObj2Key1": {
        "Obj2NestedKey1": 1,
        "obj2NestedKey2": "two"
      },
      "testObj2key2": "2"
    }
  ],
  "messageImportance": "low"
}
```

### DIFF
--- a/swift-sdk/Internal/EmbeddedMessagingManager.swift
+++ b/swift-sdk/Internal/EmbeddedMessagingManager.swift
@@ -45,7 +45,7 @@ class EmbeddedMessagingManager: NSObject, IterableEmbeddedMessagingManagerProtoc
     public func track(click message: IterableEmbeddedMessage) {
         apiClient.track(embeddedMessageClick: message)
 //        IterableAPI.track(event: "embedded-messaging", dataFields: ["name": "click",
-//                                                                    "messageId": message.metadata.id])
+//                                                                    "messageId": message.metadata.messageId])
     }
     
 //    public func track(dismiss message: IterableEmbeddedMessage) {
@@ -55,7 +55,7 @@ class EmbeddedMessagingManager: NSObject, IterableEmbeddedMessagingManagerProtoc
     public func track(impression message: IterableEmbeddedMessage) {
 //        apiClient.track(embeddedMessageImpression: message)
         IterableAPI.track(event: "embedded-messaging", dataFields: ["name": "impression",
-                                                                    "messageId": message.metadata.id])
+                                                                    "messageId": message.metadata.messageId])
     }
     
     public func track(embeddedSession: IterableEmbeddedSession) {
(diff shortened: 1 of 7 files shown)


swift-sdk/Internal/EmbeddedMessagingManager.swift
```diff
@@ -157,7 +157,7 @@ class EmbeddedMessagingManager: NSObject, IterableEmbeddedMessagingManagerProtoc
             apiClient.track(embeddedMessageReceived: message)
 //            IterableAPI.track(event: "embedded-messaging",
 //                              dataFields: ["name": "received",
-//                                           "messageId": message.metadata.id]
+//                                           "messageId": message.metadata.messageId]
```
evantk91, line 160: Seems like this section needs to be cleaned up on a separate PR. Looks like a function was added in the impression tracking PR.

swift-sdk/Internal/EmbeddedMessagingSerialization.swift
```diff
@@ -4,6 +4,38 @@
...
+                dictionary[key] = val.value
+            }
+            value = dictionary
+        } else {
+            throw DecodingError.dataCorruptedError(in: container, debugDescription: "Container contains an unexpected type")
+        }
+    }
+}
+
+
```
evantk91, line 38: This is great! But I thought we were just passing along the custom payload as a JSON. But, if this is the route we want to go, we can do something similar in Android.

swift-sdk/IterableEmbeddedMessage.swift
```diff
@@ -61,5 +62,10 @@ extension IterableEmbeddedMessage {
             public let type: String
             public let data: String?
         }
+        
+        public struct EmbeddedMessageElementsDefaultAction: Codable {
+            public let type: String
+            public let data: String?
+        }
```
evantk91, line 69: Nice I forgot about this change. good catch!

evantk91 (approved):
Looking good. Just added some notes for future development. Kind of a cursory review just making sure everything makes sense. Please confirm works with BE before merging. 👍

swift-sdk/Internal/EmbeddedMessagingSerialization.swift
```diff
@@ -4,6 +4,38 @@
 
 import Foundation
 
+struct AnyDecodable: Decodable {
```
Ayyanchira, line 7: Not sure if there should be a generic `AnyDecodable` struct. InApps also supports custom payload but seem to have another approach to it.

Ayyanchira (commented):
Left a comment.

PR #648, opened 2023-06-14
[MOB-6356] Use email or uid when making requests to api

anlinguist:
## 🔹 Jira Ticket(s)

* [MOB-6356](https://iterable.atlassian.net/browse/MOB-6356)

## ✏️ Description

This PR enables UID in embedded message requests. If the user is signed in via email, it will use email. If the user is signed in via uid, it will use uid when making requests to the API.

[MOB-6356]: https://iterable.atlassian.net/browse/MOB-6356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/RequestCreator.swift
+++ b/swift-sdk/Internal/RequestCreator.swift
@@ -436,7 +436,7 @@ struct RequestCreator {
             args[JsonKey.Embedded.packageName] = packageName
         }
         
-        addUserKey(intoDict: &args)
+        setCurrentUser(inDict: &args)
         
         return .success(.get(createGetRequest(forPath: Const.Path.getEmbeddedMessages, withArgs: args as! [String: String])))
     }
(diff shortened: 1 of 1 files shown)


swift-sdk/Internal/RequestCreator.swift
```diff
@@ -546,7 +546,7 @@ struct RequestCreator {
 
         var body = [AnyHashable: Any]()
         
-        addUserKey(intoDict: &body)
+        setCurrentUser(inDict: &body)
```
evantk91, line 550: Seems weird that there is an impression request and a embedded session request. But, thats not dealing with this change.

anlinguist: yes, I think it's legacy. seems good for a refactor stage later

PR #664, opened 2023-07-27
[MOB - 6583] - No Offline Components when Online-Only

Ayyanchira:
## 🔹 Jira Ticket(s)

* [MOB-6583](https://iterable.atlassian.net/browse/MOB-6583)

## ✏️ Description

> 1. Turned the default value to false instead of true. Only once the response comes back as true, the value will be set to true. Subsequent launches will respect whats stored after that. There will not be a toggling happening every now and then. Once the response from server is received as true, its stored in the sdk.
> 2. Create RequestHandler initializes both online and offline request Handler irrespective of the feature enabled or disabled. Added lazy var for offlineprocessor and initializing offline componenents only when offlineMode is true.

This change will make sure that components like TaskSchedular and HealthMonitor will not be even initialized unless offlineMode is true. However, this change does not engage offlineMode immediately after the response is received as true as creation of request handler has already passed and only a fresh launch of app can then go through offline request processor.

Additional changes are needed to make sure offlineMode when accessed, initializes its component and not when requestHandler is initialized. That will keep the offline component encapsulated within offline flow.

[MOB-6583]: https://iterable.atlassian.net/browse/MOB-6583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk.xcodeproj/xcshareddata/xcschemes/swift-sdk.xcscheme
+++ b/swift-sdk.xcodeproj/xcshareddata/xcschemes/swift-sdk.xcscheme
@@ -210,6 +210,11 @@
                BlueprintName = "offline-events-tests"
                ReferencedContainer = "container:swift-sdk.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "RequestHandlerTests/testFeatureFlagTurnOnOfflineMode()">
+               </Test>
+            </SkippedTests>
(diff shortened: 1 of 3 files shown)


davidtruong (approved):
Reviewed, good to go once the tests pass

PR #671, opened 2023-08-24
Add `onLogoutPreviousUserFailed` delegate to the `IterableAuthDelegate`

woosiki-prestolabs:
## 🔹 Jira Ticket(s)
I can not make Jira ticket for this because I don't have permission to access Jira.

## ✏️ Description
I am a iOS software engineer at PrestoLabs and was asked by Daryl Greaves to make a PR what I requested specifications.
I requested the specifications below but I only made this PR for 3rd one.

1. Can I use disableDeviceForAllUsers when I set autoPushRegistration as true? 
- I got a response that I can use disableDeviceForAllUsers.
2. Could you make a patch that retry register api when it failed by network error?
- I found a onFailure callback when register api was failed.
- IterableAPI.register(token: Data, , onSuccess: OnSuccessHandler? = nil, onFailure: OnFailureHandler? = nil)
3. Could you make a patch that notify when unregister device failed by 401 error?

### DIFF
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -506,7 +506,9 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         }
         
         if config.autoPushRegistration {
-            disableDeviceForCurrentUser()
+            disableDeviceForCurrentUser(onFailure:  { [weak self] reason, data in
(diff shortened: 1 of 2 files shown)


woosiki-prestolabs:
I have a question about function of `disableDeviceForAllUsers`.
I call this when jwt token already disabled, but it still fail as `{"msg":"JWT Authorization header is not set","code":"BadAuthorizationHeader","params":null}`.
Did you know how I can disable device for all users?
스크린샷 2023-08-30 오후 5 55 12

PR #677, opened 2023-10-12
Support 'merge' parameter in updateEmail API call

MichaelOrmondeHostelworld:
## 🔹 Jira Ticket(s)

* None: is support for an endpoint parameter based on ongoing Iterable implementation

## ✏️ Description

 * Add support for the "merge" parameter in the [https://api.iterable.com/api/docs#users_updateEmail](updateEmail) API endpoint
 * This parameter is not mentioned in the documentation link above, but is available on the endpoint upon activation
 * Retain support for func calls without "merge" parameter for both backwards compatibility and Obj-C interoperability
 * Relevant unit tests added

### DIFF
--- a/swift-sdk/Constants.swift
+++ b/swift-sdk/Constants.swift
@@ -96,6 +96,7 @@ enum JsonKey {
     static let currentEmail = "currentEmail"
     static let currentUserId = "currentUserId"
     static let newEmail = "newEmail"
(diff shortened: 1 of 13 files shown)


MichaelOrmondeHostelworld:
On the code climate issues, all appear to be false positives? As comparing protocols and their implementations unless I'm reading incorrectly

PR #687, opened 2023-11-07
[MOB-7252] sets up click handling for external/deep linking

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-7252](https://iterable.atlassian.net/browse/MOB-7252)

## ✏️ Description

* sets up the handleEmbeddedClick function which can be called app side where a click is handled
* sets up functionality for validating the URL and calling the `urlHandler` set up on the app side in the `urlDelegate` in the `config` 
* Adds embedded manager reset and associated unit tests

[MOB-7252]: https://iterable.atlassian.net/browse/MOB-7252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


swift-sdk/Internal/IterableEmbeddedManager.swift
```diff
@@ -43,6 +43,13 @@ class IterableEmbeddedManager: NSObject, IterableEmbeddedManagerProtocol {
     public func syncMessages(completion: @escaping () -> Void) {
         retrieveEmbeddedMessages(completion: completion)
     }
+    
+    public func handleEmbeddedClick(action: IterableAction) {
+        
+        for listener in listeners.allObjects {
+            listener.onTapAction(action: action)
+        }
+    }
```
bradumbaugh, line 52: @evantk91 do we need to track a click event here?

evantk91: No right now that is handled app side

evantk91: Moving it into the handler would make implementation easier. I will do that.

swift-sdk/EmbeddedHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+        case .applewebdata:
+            ITBError("Request url contains an invalid scheme: \(url)")
+            guard let urlPath = getUrlPath(url: url) else {
+                return nil
+            }
+            return .localResource(name: urlPath)
+        case .iterable:
+            return .iterableCustomAction(name: dropScheme(urlString: url.absoluteString, scheme: scheme.rawValue))
+        case .action, .itbl:
+            return .customAction(name: dropScheme(urlString: url.absoluteString, scheme: scheme.rawValue))
```
bradumbaugh, line 32: @evantk91 Is it possible here that the the action depends on the specific scheme that prefixes it? Maybe there are multiple different schemes supported by the app? Not sure it's safe to drop it?

swift-sdk/EmbeddedHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+                return scheme
+            } else {
+                return .other
+            }
+        }
+    }
+
+    // returns everything other than scheme, hostname and leading slashes
+    // so scheme://host/path#something => path#something
+    private static func getUrlPath(url: URL) -> String? {
```
bradumbaugh, line 60: I wonder if we could use [`URLComponents`](https://developer.apple.com/documentation/foundation/urlcomponents) for this?

bradumbaugh (commented):
@evantk91 Hey Evan, not sure I'm the right person to approve this PR, but left a couple of comments. Three questions, too:

1 - Where would the client programmer call the method to track a click event?

2 - After a click, how would the client programmer know to update the UI that displayed the message?

3 - I think we're not supporting custom actions for now, is that right?

swift-sdk/Internal/IterableEmbeddedManager.swift
```diff
@@ -43,8 +50,58 @@ class IterableEmbeddedManager: NSObject, IterableEmbeddedManagerProtocol {
     public func syncMessages(completion: @escaping () -> Void) {
         retrieveEmbeddedMessages(completion: completion)
     }
+    
+    public func handleEmbeddedClick(message: IterableEmbeddedMessage?, buttonIdentifier: String?, clickedUrl: String) {
```
vbabenkoru, line 54: Why is `message` optional?

swift-sdk/Internal/IterableEmbeddedManager.swift
```diff
@@ -43,8 +50,58 @@ class IterableEmbeddedManager: NSObject, IterableEmbeddedManagerProtocol {
...
+        }
+
+        if let url = URL(string: clickedUrl) {
+            handleClick(clickedUrl: url, forMessage: message)
+        } else {
+            print("Invalid URL: \(clickedUrl)")
+        }
+    }
+    
+    private func handleClick(clickedUrl url: URL, forMessage message: IterableEmbeddedMessage) {
```
vbabenkoru, line 68: Why is it a separate method?

swift-sdk/Internal/IterableEmbeddedManager.swift
```diff
@@ -43,8 +50,58 @@ class IterableEmbeddedManager: NSObject, IterableEmbeddedManagerProtocol {
...
+        
+        handleUrl(url: url.absoluteString)
+    }
+    
+    
+    private func createAction(fromUrlOrAction url: String) -> IterableAction? {
```
vbabenkoru, line 74: Since this is copied from the in-app manager code, maybe it'd be better to move this stuff to some kind of a helper class to reduce duplication.

vbabenkoru:
Could you add some unit tests for this?

evantk91:
> Could you add some unit tests for this?

I will add as a fast follow on a separate PR since we are targeting testing by next wednesday.

evantk91:
> > Could you add some unit tests for this?
> 
> I will add as a fast follow on a separate PR since we are targeting testing by next wednesday.

This functionality is actually covered by `ActionRunnerTests.swift`. The added functionality for embedded just processes the click information and passes it to the same action runner that in-app messages use.

PR #704, opened 2023-12-14
[MOB-7458] - Aligning iOS with Android

Ayyanchira:
## 🔹 Jira Ticket(s)

* [MOB-7458](https://iterable.atlassian.net/browse/MOB-7458)

## ✏️ Description

Config object
Enum for type
Removal of ViewDelegate and its methods

[MOB-7458]: https://iterable.atlassian.net/browse/MOB-7458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/IterableEmbeddedManager.swift
+++ b/swift-sdk/Internal/IterableEmbeddedManager.swift
@@ -44,6 +44,7 @@ class IterableEmbeddedManager: NSObject, IterableEmbeddedManagerProtocol {
         
         super.init()
         addForegroundObservers()
+        syncMessages { print("Retrieving embedded message")}
     }
     
     var onDeinit: (() -> Void)?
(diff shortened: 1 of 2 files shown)


evantk91 (approved):
LGTM at least for today's bug bash.

A few points we need to address before release:

- notification default is not aligned with Android. The background and border colors differ from the card and banner styles.
- the second button is defaulting to this weird block button. it should be a button with a transparent background and the same color text as the first button. See the figma.
- the second button background and border color should be configurable. I will add this to the android side.
- the card view type should collapse when the image is not provided to align with the android side and the discussion I had with design (stephanie)

Ayyanchira:
Will bug bash with this branch so we can work on this before merging into embedded

evantk91:
> LGTM at least for today's bug bash.
> 
> A few points we need to address before release:
> 
> * notification default is not aligned with Android. The background and border colors differ from the card and banner styles.
> * the second button is defaulting to this weird block button. it should be a button with a transparent background and the same color text as the first button. See the figma.
> * the second button background and border color should be configurable. I will add this to the android side.
> * the card view type should collapse when the image is not provided to align with the android side and the discussion I had with design (stephanie)

* point 2 addressed

evantk91 (approved):
LGTM. merging with base feature branch.

PR #728, opened 2024-01-16
Sync events if setUser is called

hani-iterable:
## ✏️ Description

> We synchronise events if setUser is called before creating a known user.


hani-iterable:
We can close this PR as it includes all the changes related to the issue. You can find the updated changes in [PR](https://github.com/Iterable/swift-sdk/pull/732).

PR #740, opened 2024-01-26
[MOB-?] Fix InAppMessage not being displayed due to correct KeyWindow not being retrieved on app that uses multiple windows

harrywombat:
## 🔹 Jira Ticket(s)

## ✏️ Description

If a project makes use of multiple Windows, then the IterableSDK can fail to retrieve the correct one, resulting in the InAppMessage failing and never being displayed.

This fix allows the currently displayed Window to be retrieved (the true keyWindow), so the InAppMessage can be displayed on the very top view controller.

It also accounts for the iOS 15 changes to windows.


jena-chakour:
Thanks for sharing this, Harry!

Ayyanchira (approved):
Looks good. Merging in a feature branch to test locally

PR #760, opened 2024-04-09
Mapping xcprivacy in podspec file

silvgabriel:
Mapping xcprivacy in podspec file

## 🔹 Jira Ticket(s)

## ✏️ Description

### DIFF
--- a/Iterable-iOS-SDK.podspec
+++ b/Iterable-iOS-SDK.podspec
@@ -23,7 +23,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = '5.3'
 
-  s.resource_bundles = {'Resources' => 'swift-sdk/Resources/**/*.{storyboard,xib,xcassets,xcdatamodeld}' }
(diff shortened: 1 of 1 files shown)


silvgabriel:
You forgot to mapping the PrivacyInfo file type in the podspec

Ayyanchira:
Thanks for the PR!

I had tested an App with SPM generated Privacy report which had data from SDK's manifest. 

@silvgabriel, does the Cocoapods based project not pull privacy manifest from SDK when generating report? If so, does this change rectify the issue?

silvgabriel:
Hi @Ayyanchira 

SPM is working well because you only map the `Resources` folder and everything we have inside this folder will appear in the project.
In the CocoaPod podspec you map the file types in the resource part and the `xcprivacy` type is missing, so the private manifest file does not appear. 

This PR fixes the CocoaPod part, check the image below, it is the `Iterable-iOS-SDK` pod installed in my project:
image

Another way to resolve this is to equalize the podspec with the SPM file by mapping just the `Resource` folder (without specifying types).

PR #221, opened 2019-11-19
[HOLD] [MOB-901] Reorganize 3rd party configs

roninopf:

PR #780, opened 2024-06-28
MOB-8822

hardikmashru:
## 🔹 Jira Ticket(s)

* [MOB-8822](https://iterable.atlassian.net/browse/MOB-8822)
* [MOB-8823](https://iterable.atlassian.net/browse/MOB-8823)

## ✏️ Description

[MOB-8822]: https://iterable.atlassian.net/browse/MOB-8822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MOB-8823]: https://iterable.atlassian.net/browse/MOB-8823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -217,6 +217,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
             if config.enableAnonTracking {
                 anonymousUserManager.trackAnonTokenRegistration(token: token.hexString())
             }
+            onFailure?("Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods", nil)
             return
         }
         
@@ -325,9 +326,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
     private func rejectWithInitializationError(onFailure: OnFailureHandler? = nil) -> Pending<SendRequestValue, SendRequestError> {
         let result = Fulfill<SendRequestValue, SendRequestError>()
         result.reject(with: SendRequestError())
-        if let _onFailure = onFailure {
-            _onFailure("Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods", nil)
-        }
+        onFailure?("Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods", nil)
         return result
     }
     
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -213,8 +213,11 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
             return
         }
         
-        if !isEitherUserIdOrEmailSet() && config.enableAnonTracking && localStorage.userIdAnnon == nil {
-            anonymousUserManager.trackAnonTokenRegistration(token: token.hexString())
+        if !isEitherUserIdOrEmailSet() && localStorage.userIdAnnon == nil {
+            if config.enableAnonTracking {
+                anonymousUserManager.trackAnonTokenRegistration(token: token.hexString())
+            }
+            return
         }
         
         hexToken = token.hexString()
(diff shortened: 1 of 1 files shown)


swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -213,8 +213,12 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
             return
         }
         
-        if !isEitherUserIdOrEmailSet() && config.enableAnonTracking && localStorage.userIdAnnon == nil {
-            anonymousUserManager.trackAnonTokenRegistration(token: token.hexString())
+        if !isEitherUserIdOrEmailSet() && localStorage.userIdAnnon == nil {
+            if config.enableAnonTracking {
+                anonymousUserManager.trackAnonTokenRegistration(token: token.hexString())
+            }
+            onFailure?("Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods", nil)
```
evantk91, line 220: this failure message is inaccurate since the message could be triggered by the anonymous user tracking being disabled and also a criteria not being matched

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -271,8 +275,11 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
                     mergeNestedObjects: Bool,
                     onSuccess: OnSuccessHandler? = nil,
                     onFailure: OnFailureHandler? = nil) -> Pending<SendRequestValue, SendRequestError> {
-        if !isEitherUserIdOrEmailSet() && config.enableAnonTracking && localStorage.userIdAnnon == nil{
-            anonymousUserManager.trackAnonUpdateUser(dataFields)
+        if !isEitherUserIdOrEmailSet() && localStorage.userIdAnnon == nil {
+            if config.enableAnonTracking {
+                anonymousUserManager.trackAnonUpdateUser(dataFields)
+            }
+            return rejectWithInitializationError(onFailure: onFailure)
```
evantk91, line 282: Remove the rejectWithInitializationError

evantk91 (commented):
Left some comments

swift-sdk/Internal/InternalIterableAPI.swift
hardikmashru: Hi @evantk91 this message is accurate because if you notice if condition message is triggered when SDK is NOT initialized. When SDK gets initialised it assigns null to localStorage.userIdAnnon anyways. 

Please let me know if you still have any doubts.

swift-sdk/Internal/InternalIterableAPI.swift
hardikmashru: if you notice we need to return the result of type Pending<SendRequestValue, SendRequestError> and that is why we cannot remove rejectWithInitializationError.

swift-sdk/Internal/InternalIterableAPI.swift
evantk91: The SDK initialization check needs to be separated and should stand alone. Right now, if anonymous user trracking is disabled, the track calls will be made regardless of if SDK is initialized. We should have a separate check that returns if SDK is not initialized.

evantk91: Also, the `localStorage.userIdAnnon` check can be removed. We want to update flow so that setUserId is called when  a known user is created. I will make a ticket for this if necessary.

hardikmashru: Hi @evantk91 If you notice line #221 it returns from there so in case SDK is not initialized track calls won't be made. We need to make API calls if localStorage.userIdAnnon is NOT NULL.

Can you please create a ticket for the new flow with an explanation so we know if removing localStorage.userIdAnnon makes sense or NOT?

evantk91: I will take a look. Can you separate out the SDK initialization check with its own return statement? We dont want the requestHandler to be called if the SDK is not initialized and anonymous tracking is disabled.

evantk91: I will create a new ticket if necessary.

evantk91 (approved):

PR #790, opened 2024-07-10
[MOB-8962] updates merge flow to align with Android

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-8962](https://iterable.atlassian.net/browse/MOB-8962)

## ✏️ Description

> This pull request makes a call to setUserId when the annoymous user is created. I also updated the flow to accommodate the default merge parameter which should be true when the merge parameter is not provided.


swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -168,18 +169,12 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
...
-               if (!isEmailOrUserId && anonUser == nil) { // Criteria is not yet met (default merge is true)
-                   return true;
-            } else if (!isEmailOrUserId && anonUser != nil) { // Criteria is met (Iterable profile created with an autogenerated identity)(default merge is true)
-                return true;
-            } else { // Current logged in user is identified (default merge is false)
-                return false;
-            }
+            return merge
+        } else {
+            return true
```
Ayyanchira, line 177: Why is this function required anymore if an `if` if the only remaining logic?

swift-sdk/IterableAPI.swift
```diff
@@ -151,8 +151,8 @@ import UIKit
         implementation?.setEmail(email, authToken: authToken, merge: merge, successHandler: successHandler, failureHandler: failureHandler)
     }
     
-    public static func setUserId(_ userId: String?, _ authToken: String? = nil, merge: Bool? = nil, _ successHandler: OnSuccessHandler? = nil, _ failureHandler: OnFailureHandler? = nil) {
-        implementation?.setUserId(userId, authToken: authToken, merge: merge,successHandler: successHandler, failureHandler: failureHandler)
+    public static func setUserId(_ userId: String?, _ authToken: String? = nil, merge: Bool? = nil, _ successHandler: OnSuccessHandler? = nil, _ failureHandler: OnFailureHandler? = nil, _ isAnon: Bool = false) {
+        implementation?.setUserId(userId, authToken: authToken, merge: merge,successHandler: successHandler, failureHandler: failureHandler, isAnon: isAnon)
```
Ayyanchira, line 155: Are these both `isAnon` and `merge` required to make the anonymous flow work correctly. setUserId method might get more complicated with additional two flags for one feature. Should developers be setting the `isAnon` to true? 
whether a user is anonymous should be something only SDK should understand and process internally.

Ayyanchira (commented):
Made some comments

swift-sdk/Internal/AnonymousUserMerge.swift
```diff
@@ -21,17 +21,20 @@ class AnonymousUserMerge: AnonymousUserMergeProtocol {
...
-                apiClient.mergeUser(sourceEmail: sourceEmail, sourceUserId: sourceUserId,  destinationEmail : destinationEmail, destinationUserId: destinationUserId).onSuccess {_ in
-                    onMergeResult(MergeResult.mergesuccessful, nil)
-                }.onError {error in
-                    print("Merge failed error: \(error)")
-                    onMergeResult(MergeResult.mergefailed, error.reason)
-                }
-            } else {
-                // this will return mergeResult true in case of anon userId doesn't exist or destinationUserIdOrEmail is nil because merge is not required
-                onMergeResult(MergeResult.mergenotrequired, nil)
+    public func tryMergeUser(sourceUserId: String?, sourceEmail: String?, destinationUserIdOrEmail: String?, isEmail: Bool, merge: Bool, onMergeResult: @escaping MergeActionHandler) {
```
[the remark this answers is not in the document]
evantk91: Yes, in the case where a user is logging out. The destination id/email is nil.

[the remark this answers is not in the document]
evantk91: In this case, tryMerge does not call the merge API and just returns the result that merge is not required.

swift-sdk/Internal/AnonymousUserMerge.swift
```diff
@@ -21,17 +21,20 @@ class AnonymousUserMerge: AnonymousUserMergeProtocol {
...
+        if (sourceUserId != nil && destinationUserIdOrEmail != nil && merge) {
```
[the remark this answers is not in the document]
evantk91: Not sure I follow this.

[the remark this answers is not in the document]
evantk91: I see. From the earlier point, the source and destination are made optional to allow for log out.

swift-sdk/Internal/InternalIterableAPI.swift
evantk91: Yeah true we can simplify and just add a conditional to the function body

swift-sdk/IterableAPI.swift
evantk91: `isAnon` is needed with the current flow since the annoymous user id is local storage is cleared if it is not provided currently. We can brainstorm a way to re-architect it correctly

evantk91: The flag is used internally. The end user will never have to set the flag. Currently it is only used internally to set an anonymous user.

evantk91: Actually it is public facing. We should have the manager call the internal iterable API.

Ayyanchira (approved):
LGTM

PR #795, opened 2024-08-01
Fix iterable SDK bug #592

### LINKED ISSUE #592: handle(universalLink) does not behave as described.
We upgrade from version 6.3.4 to version 6.4.9 of the Iterable SDK, and we ran into an issue where our deeplink routes are being routed twice. 

It was discovered that the reason for this is the call to "handle(universalLink url: URL)" is now returning false where it used to return true for the same deeplink. 

Upon closer examination, the reason seems to be that as per the documentation on this function, it says `Returns: true if it is an Iterable link, or the value returned from IterableURLDelegate otherwise`. However, the call makes it through to the URL Delegate, which returns true. But the call to handle(universalLink url: URL) returns false. This seems to be because the call inside DeepLinkManager-> handleUniversalLink went from this:
    
Screenshot 2023-02-22 at 11 21 41 AM

to this
    
Screenshot 2023-02-22 at 11 22 10 AM

As you can see it is no longer capturing the result from ActionRunner.execute and always returns `false` no matter what the URLDelegate returns.

scottasoutherland:
## ✏️ Description

This is a pull request to fix #592 

I have manually verified it fixes the issue still present  in 6.5.4

### DIFF
--- a/swift-sdk/Internal/DeepLinkManager.swift
+++ b/swift-sdk/Internal/DeepLinkManager.swift
@@ -43,17 +43,18 @@ class DeepLinkManager: NSObject {
             // Always return true for deep link
             return (true, pending)
         } else {
+            var result: Bool = false
             if let action = IterableAction.actionOpenUrl(fromUrlString: url.absoluteString) {
                 let context = IterableActionContext(action: action, source: .universalLink)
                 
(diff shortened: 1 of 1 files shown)


davidtruong (approved):
Thanks, Change looks good to me.

PR #796, opened 2024-08-05
[MOB-9247] - Bug fix for no push on APNS SANDBOX

Ayyanchira:
## 🔹 Jira Ticket(s)

* [MOB-9247](https://iterable.atlassian.net/browse/MOB-9247)

### Logic explanation
SDK checks if the target is a simulator
if so > APNS SANDBOX
if not > read the file 'embedded.provision'
Check if certain value exist to mark it as APNSSANDBOX If not, default it to PRODUCTION APNS.

Using Xcode 16 beta with iOS 18 beta, we discovered that file reading was failing with normal ASCII encoding.

### Diff explanation -
File was able to type to data easily.
Once the data was available, now we try to read the file with all encoding available to us now. If none of them works, we print ITBLError. The first one to decode the file proceeds with scan operation - finding the right tags and keywords.

- Tested with Xcode GA and Beta and iOS 17 and iOS 18 devices

[MOB-9247]: https://iterable.atlassian.net/browse/MOB-9247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/APNSTypeChecker.swift
+++ b/swift-sdk/Internal/APNSTypeChecker.swift
@@ -59,12 +59,39 @@ struct APNSTypeChecker: APNSTypeCheckerProtocol {
     }
     
     static func readMobileProvision(fromPath path: String) -> [AnyHashable: Any] {
-        guard let asciiString = try? String(contentsOfFile: path, encoding: .ascii) else {
+        guard let data = try? Data(contentsOf: URL(fileURLWithPath: path)) else {
             ITBError("Could not read file: \(path)")
             return [:]
         }
+
+        let encodings: [String.Encoding] = [
+            .ascii,
+            .utf8,
+            .utf16,
+            .utf16BigEndian,
+            .utf16LittleEndian,
+            .utf32,
+            .utf32BigEndian,
+            .utf32LittleEndian,
+            .isoLatin1,
+            .macOSRoman
+        ]
         
-        guard let propertyListString = scan(string: asciiString, begin: "<plist", end: "</plist>") else {
+        var asciiString: String?
+        
+        for encoding in encodings {
+            if let string = String(data: data, encoding: encoding) {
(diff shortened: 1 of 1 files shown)


swift-sdk/Internal/APNSTypeChecker.swift
```diff
@@ -59,12 +59,39 @@ struct APNSTypeChecker: APNSTypeCheckerProtocol {
...
+            .utf8,
+            .utf16,
+            .utf16BigEndian,
+            .utf16LittleEndian,
+            .utf32,
+            .utf32BigEndian,
+            .utf32LittleEndian,
+            .isoLatin1,
+            .macOSRoman
+        ]
```
vbabenkoru, line 78: Which one actually works? Do we need to try all?

Ayyanchira: `.isoLatin1` worked for iOS 18. Ascii works with pre iOS 18 devices. 
You think I should narrow down to two that I know works instead of this?

omnicgdave: Did you need to update this with the appropriate encodings?

virusman: @Ayyanchira I think we should at least put the verified working ones at the top, to reduce the risk of a situation where the data can be decoded with an encoding, but is then decoded incorrectly, and the string is mangled/can't be interpreted.

virusman: Trying everything and continuing with the first thing that decodes it doesn't 100% guarantee that the decoded value would be valid.

vbabenkoru (approved):

PR #814, opened 2024-08-23
[MOB-9421] make tryMergeUser private

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-9421](https://iterable.atlassian.net/browse/MOB-9421)

## ✏️ Description

> This pull request makes tryMergeUser private and updates tryMergeUser to only take in an anonymous user as the source user id.

[MOB-9421]: https://iterable.atlassian.net/browse/MOB-9421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/AnonymousUserMerge.swift
+++ b/swift-sdk/Internal/AnonymousUserMerge.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 protocol AnonymousUserMergeProtocol {
-    func tryMergeUser(sourceUserId: String?, destinationUserIdOrEmail: String?, isEmail: Bool, merge: Bool, onMergeResult: @escaping MergeActionHandler)
+    func tryMergeUser(anonymousUserId: String?, destinationUser: String?, isEmail: Bool, shouldMerge: Bool, onMergeResult: @escaping MergeActionHandler)
 }
 
 class AnonymousUserMerge: AnonymousUserMergeProtocol {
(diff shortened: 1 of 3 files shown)


swift-sdk/Internal/AnonymousUserMerge.swift
```diff
@@ -21,12 +21,12 @@ class AnonymousUserMerge: AnonymousUserMergeProtocol {
...
-        if (sourceUserId != nil && destinationUserIdOrEmail != nil && merge) {
-            let destinationEmail = isEmail ? destinationUserIdOrEmail : nil
-            let destinationUserId = isEmail ? nil : destinationUserIdOrEmail
+    func tryMergeUser(anonymousUserId: String?, destinationUser: String?, isEmail: Bool, shouldMerge: Bool, onMergeResult: @escaping MergeActionHandler) {
+        if (anonymousUserId != nil && destinationUser != nil && shouldMerge) {
+            let destinationEmail = isEmail ? destinationUser : nil
+            let destinationUserId = isEmail ? nil : destinationUser
             
-            apiClient.mergeUser(sourceEmail: sourceEmail, sourceUserId: sourceUserId,  destinationEmail : destinationEmail, destinationUserId: destinationUserId).onSuccess {_ in
+            apiClient.mergeUser(sourceEmail: nil, sourceUserId: anonymousUserId,  destinationEmail: destinationEmail, destinationUserId: destinationUserId).onSuccess {_ in
```
Ayyanchira, line 29: Also now that we know sourceEmail will always be nil, do we need to pass in anonymousUserID explicitly? SDK should know internally what userID it has.

evantk91: Gotcha. Updated so that the local storage is referenced in the AnonymousUserMerge class.

Ayyanchira (approved):
Looking great!

PR #825, opened 2024-09-17
Take into consideration not determined status into checking notifications access status

piotrtorczynski:
## ✏️ Description

```swift
let config = IterableConfig()
config.autoPushRegistration = false
IterableAPI.initialize(apiKey: apiKey, launchOptions: launchOptions, config: config)
```

If you take into consideration that the user has `autoPushRegistration` set to false, the SDK will return `notificationsEnabled` set to true due to this bug. If `autoPushRegistration` is set to false, the user often has `UNAuthorizationStatus` set to `.notDetermined`.  

This change fixes this issue so that by default SDK should return the proper value


sumeruchat (approved):
That makes sense. Thanks for the PR.

PR #838, opened 2024-10-08
[MOB-9727] additional complex criteria test

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-9727](https://iterable.atlassian.net/browse/MOB-9727)

## ✏️ Description

> This pull request adds an additional test case for complex criteria #3 test so that both possible passing events are covered. I will add this case to Android too for parity.

[MOB-9727]: https://iterable.atlassian.net/browse/MOB-9727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ayyanchira (approved):
LGTM. Considering that 
Not > will have additions to its segmentation when we compliment it with AND and OR. 
at the same time, we can do the exclusion from the parent NOR or (Cant have) by adding a NOR or NOT.. 
feels complex but makes sense..

PR #852, opened 2024-10-24
Evan/mob 9946

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-9646](https://iterable.atlassian.net/browse/MOB-9646)

## ✏️ Description

> This pull request is an extension of the other PR where I modify some of the public methods and remove some of the initialization checks.


swift-sdk/IterableAPI.swift
```diff
@@ -332,9 +336,8 @@ import UIKit
                                   mergeNestedObjects: Bool,
                                   onSuccess: OnSuccessHandler? = nil,
                                   onFailure: OnFailureHandler? = nil) {
-        guard let implementation, implementation.isSDKInitialized() else { return }
         
-        implementation.updateUser(dataFields,
+        implementation?.updateUser(dataFields,
```
Ayyanchira, line 340: why is here ->  implementation?.callingMethod -> lets call approach A
and above its guart let unwrap the variable and call the method. approach B

overall looks like we are doing B for disableDeviceApis and A for update and track in this PR.

evantk91: We dont want to gate the track calls since they need to be called for AUA

PR #866, opened 2024-11-25
[MOB-10104] Fix wakey notification tracking event on did finish launch with options 

joaodordio:
## 🔹 Jira Ticket(s)

* [MOB-10104](https://iterable.atlassian.net/browse/MOB-10104)

## ✏️ Description

When a Push notification with the Wakey feature tuned on — `"content-available": 1` — the SDK was tracking a pushOpen event incorrectly.

Now we actively check if it's a wakey push notification and prevent that tracking — unless the user actually taps on the notification and opens the app.

[MOB-10104]: https://iterable.atlassian.net/browse/MOB-10104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -706,7 +706,16 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         guard let launchOptions = launchOptions else {
             return
         }
+        
         if let remoteNotificationPayload = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] {
+            
+            if let aps = remoteNotificationPayload["aps"] as? [String: Any],
+               let contentAvailable = aps["content-available"] as? Int,
+               contentAvailable == 1 {
+                ITBInfo("Received push notification with wakey content-available flag")
+                return
+            }
+            
             if let _ = IterableUtil.rootViewController {
                 // we are ready
                 IterableAppIntegration.implementation?.performDefaultNotificationAction(remoteNotificationPayload)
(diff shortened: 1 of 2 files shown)


swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -706,7 +706,16 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
...
             return
         }
+        
         if let remoteNotificationPayload = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] {
+            
+            if let aps = remoteNotificationPayload[Const.RemoteNotification.aps] as? [String: Any],
+               let contentAvailable = aps[Const.RemoteNotification.contentAvailable] as? Int,
+               contentAvailable == 1 {
+                ITBInfo("Received push notification with wakey content-available flag")
+                return
```
Ayyanchira, line 716: Looks like this can indeed stop the further calling of performing defaultNotification actions. But will it prevent any background processing the app needed to do?

Ayyanchira: Also thinking if we return we will miss these events and if these events needed a dedicated tracking. Something like - background push delivery events?

Ayyanchira (commented):
Shared some thoughts

swift-sdk/Internal/InternalIterableAPI.swift
joaodordio: > will it prevent any background processing the app needed to do

Not really I think. The `handle(launchOptions:)`  only tracks the push default action and performs the default action — which should not happen on the background delivery of the push.

As to the second question, we can add that functionality afterwards. Not sure if we are tracking those on the other platforms. What do you think?

joaodordio: @Ayyanchira gentle ping on this one!

Ayyanchira (approved):
Looking good!

PR #880, opened 2025-01-03
MOB-9233 Support for In App JSON Only Messages

sumeruchat:
## 🔹 Jira Ticket(s)

* [MOB-9233](https://iterable.atlassian.net/browse/MOB-9233)

## ✏️ Description

Add support for json only in app messages

[MOB-9233]: https://iterable.atlassian.net/browse/MOB-9233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/in-app/InAppManager+Functions.swift
+++ b/swift-sdk/Internal/in-app/InAppManager+Functions.swift
@@ -6,7 +6,7 @@ import Foundation
 
 enum MessagesProcessorResult {
     case show(message: IterableInAppMessage, messagesMap: OrderedDictionary<String, IterableInAppMessage>)
-    case noShow(messagesMap: OrderedDictionary<String, IterableInAppMessage>)
+    case noShow(message: IterableInAppMessage?, messagesMap: OrderedDictionary<String, IterableInAppMessage>)
 }
 
 struct MessagesProcessor {
@@ -32,9 +32,9 @@ struct MessagesProcessor {
             return processMessages()
         case let .skipAndConsume(message):
             updateMessage(message, didProcessTrigger: true, consumed: true)
-            return processMessages()
+            return .noShow(message: message, messagesMap: messagesMap)
         case .none, .wait:
-            return .noShow(messagesMap: messagesMap)
+            return .noShow(message: nil, messagesMap: messagesMap)
         }
     }
     
--- a/swift-sdk/Internal/in-app/InAppManager+Functions.swift
+++ b/swift-sdk/Internal/in-app/InAppManager+Functions.swift
@@ -30,6 +30,9 @@ struct MessagesProcessor {
         case let .skip(message):
             updateMessage(message, didProcessTrigger: true)
             return processMessages()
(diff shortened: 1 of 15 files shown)


swift-sdk/Internal/in-app/InAppManager+Functions.swift
```diff
@@ -59,7 +63,11 @@ struct MessagesProcessor {
         
         ITBDebug("isOkToShowNow")
         
-        if inAppDelegate.onNew(message: message) == .show {
+        let returnValue = inAppDelegate.onNew(message: message)
+        if message.isJsonOnly {
+            return .skipAndConsume(message)
+        }
+        if returnValue == .show {
```
joaodordio, line 70: Maybe we don't need to hold the inAppDelegate onNew value in a var.

```suggestion
        if message.isJsonOnly {
            return .skipAndConsume(message)
        }
        if inAppDelegate.onNew(message: message) == .show {
```

swift-sdk/Internal/in-app/InAppManager.swift
```diff
@@ -316,6 +323,10 @@ class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
                               callback: ITBURLCallback? = nil) {
         ITBInfo()
         
+        if message.isJsonOnly {
+            return
+        }
+
```
joaodordio, line 329: Just personal preference to keep up with the style of the rest of the func.
```suggestion
        guard !message.isJsonOnly else {
            return
        }
        
```

swift-sdk/Internal/in-app/InAppMessageParser.swift
```diff
@@ -81,41 +81,78 @@ struct InAppMessageParser {
             return .failure(.parseFailed(reason: "no messageId", messageId: nil))
         }
         
-        guard let contentDict = json[JsonKey.InApp.content] as? [AnyHashable: Any] else {
-            return .failure(.parseFailed(reason: "no content in json payload", messageId: messageId))
-        }
-        
-        let content: IterableInAppContent
+        let jsonOnly = (json[JsonKey.InApp.jsonOnly] as? Int ?? 0) == 1
```
joaodordio, line 84: I would add that comment just to clarify what's happening. the one liner is hard to read in one pass.
```suggestion
        // Check if the jsonOnly key is present and is set to 1 (true)
        let jsonOnly = (json[JsonKey.InApp.jsonOnly] as? Int ?? 0) == 1
```

joaodordio (approved):
Left a couple of suggestions.
Can you also please see why the CI is not running?

swift-sdk/Internal/in-app/InAppManager+Functions.swift
sumeruchat: well we have to call onNew on a json only message before returning so thats why we call and store the return value and then ignore it for json only

PR #887, opened 2025-01-17
[MOB-10626] removes disable push call

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-10626](https://iterable.atlassian.net/browse/MOB-10626)

## ✏️ Description

> This pull request removes the disable device call from the notification state tracking.

[MOB-10626]: https://iterable.atlassian.net/browse/MOB-10626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (approved):

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -741,11 +736,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
             
             if self.isEitherUserIdOrEmailSet() {
                 if hasStoredPermission && (storedEnabled != systemEnabled) {
-                    if !systemEnabled {
-                        self.disableDeviceForCurrentUser()
-                    } else {
-                        self.notificationStateProvider.registerForRemoteNotifications()
-                    }
+                    self.notificationStateProvider.registerForRemoteNotifications()
```
sumeruchat, line 739: I dont know if this will do this because when notifications are disabled the system will not give us a token. Instead we probably have to call registerDevice with the previous token and (notificationsEnabled) set to false

sumeruchat: Lets test and figure out first

sumeruchat: I just tested this and iOS does give us a token even when the system setting is disabled.

PR #897, opened 2025-02-05
MOB-9982 Embedded messaging crash when image fails to load

joaodordio:
## 🔹 Jira Ticket(s)

* [MOB-9982](https://iterable.atlassian.net/browse/MOB-9982)

## ✏️ Description

Instead of force unwrapping a potential nil value stored on the imgView and cardImageView components, we now use optionals to prevent a crash.

[MOB-9982]: https://iterable.atlassian.net/browse/MOB-9982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/ui-components/uikit/IterableEmbeddedView.swift
+++ b/swift-sdk/ui-components/uikit/IterableEmbeddedView.swift
@@ -9,7 +9,7 @@ import Foundation
 import UIKit
 
 @IBDesignable
-public class IterableEmbeddedView:UIView {
+public class IterableEmbeddedView: UIView {
     
     /// Set background color of view in container view.
     @IBOutlet weak public var contentView: UIView!
@@ -290,10 +290,6 @@ public class IterableEmbeddedView:UIView {
                 cardImageTopConstraint.isActive = false
                 titleToTopConstraint.isActive = true
                 titleToTopConstraint?.isActive = true
-                
-                // Remove cardImageView from its superview and release it
-                cardImageView?.removeFromSuperview()
-                cardImageView = nil
             }
         case .banner:
             imgView?.isHidden = EMimage == nil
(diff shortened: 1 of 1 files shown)


swift-sdk/ui-components/uikit/IterableEmbeddedView.swift
```diff
@@ -32,8 +32,8 @@ public class IterableEmbeddedView:UIView {
     @IBOutlet weak var horizontalButtonStackViewSpacer: UIView!
     
     /// IterableEmbeddedView Image View.
-    @IBOutlet weak public var imgView: UIImageView!
-    @IBOutlet weak public var cardImageView: UIImageView!
+    @IBOutlet weak public var imgView: UIImageView?
```
Ayyanchira, line 35: This is the fix right there! Good find! We should always expect image to not come just in case..

swift-sdk/ui-components/uikit/IterableEmbeddedView.swift
```diff
@@ -274,46 +274,46 @@ public class IterableEmbeddedView:UIView {
...
+            } else {
+                // Hide cardImageView and deactivate its constraints
+                cardImageView?.isHidden = true
                 cardImageTopConstraint.isActive = false
                 titleToTopConstraint.isActive = true
-                cardImageTopConstraint?.isActive = false
                 titleToTopConstraint?.isActive = true
+                
+                // Remove cardImageView from its superview and release it
+                cardImageView?.removeFromSuperview()
```
Ayyanchira, line 295: does it need to be removed ? just hidden should suffice right?

Ayyanchira (commented):
Gave it a first pass

swift-sdk/ui-components/uikit/IterableEmbeddedView.swift
joaodordio: That's a good catch! We shouldn't be removing and adding this to the view hierarchy!

PR #898, opened 2025-02-12
[MOB-8113] adds placement ids array to sync messages

evantk91:
## 🔹 Jira Ticket(s)

* [MOB-8113](https://iterable.atlassian.net/browse/MOB-8113)

## ✏️ Description

> This pull request adds optional parameter for placement ids to sync messages to make use of optional parameter on get messages API.

[MOB-8113]: https://iterable.atlassian.net/browse/MOB-8113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


swift-sdk/Internal/api-client/Request/RequestCreator.swift
```diff
@@ -424,6 +424,10 @@ struct RequestCreator {
     // MARK: - Embedded Messaging Request Calls
     
     func createGetEmbeddedMessagesRequest() -> Result<IterableRequest, IterableError> {
+        return createGetEmbeddedMessagesRequest(placementIds: [])
+    }
+    
+    func createGetEmbeddedMessagesRequest(placementIds: [Int]) -> Result<IterableRequest, IterableError> {
```
joaodordio:
```suggestion
    func createGetEmbeddedMessagesRequest(placementIds: [Int]? = []) -> Result<IterableRequest, IterableError> {
```


I would just update the method here to take in a default value of an empty array instead of cascading the logic.


The only other place where `createGetEmbeddedMessagesRequest()` is called is at 

`let result = createRequestCreator().flatMap { $0.createGetEmbeddedMessagesRequest(placementIds: placementIds ?? []) }` 

which can also be simplified to 

`let result = createRequestCreator().flatMap { $0.createGetEmbeddedMessagesRequest(placementIds: placementIds) }` 

because `createGetEmbeddedMessagesRequest` gets an optional array now

swift-sdk/Internal/api-client/Request/RequestCreator.swift
```diff
@@ -437,6 +441,10 @@ struct RequestCreator {
             args[JsonKey.Embedded.packageName] = packageName
         }
         
+        if !placementIds.isEmpty {
```
joaodordio:
```suggestion
        if let placementIds = placementIds,
           !placementIds.isEmpty {
```

joaodordio (commented):
little changes

joaodordio (approved):

PR #907, opened 2025-03-14
[MOB-11000] CI Reporting

sumeruchat:
Create nice reporting with xcresulttool


evantk91 (approved):
Nice one! I skipped some tests on a separate PR and purposely failed some. Looks good. We may want to show the skipped tests in the report, just a thought. The 2 skips that I added were shown but not the other 13 it seems.
Screenshot 2025-03-25 at 3 15 24 PM

joaodordio (approved):

PR #938, opened 2025-08-13
[MOB-XXXX] Evan/refactor

evantk91:
## 🔹 Jira Ticket(s)

## ✏️ Description

### DIFF
--- a/swift-sdk/Internal/UnknownUserManager.swift
+++ b/swift-sdk/Internal/UnknownUserManager.swift
@@ -30,6 +30,16 @@ public class UnknownUserManager: UnknownUserManagerProtocol {
     private(set) var lastCriteriaFetch: Double = 0
     private var isCriteriaMatched = false
 
+    // MARK: - Pending Consent Tracking
+    /// Holds consent data that should be sent once user creation or registration is confirmed
+    private struct PendingConsentData {
+        let consentTimestamp: Int64
+        let email: String?
+        let userId: String?
+        let isUserKnown: Bool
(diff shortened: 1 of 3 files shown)


swift-sdk/Internal/UnknownUserManager.swift
```diff
@@ -87,6 +97,60 @@ public class UnknownUserManager: UnknownUserManagerProtocol {
...
+
+        IterableAPI.implementation?.apiClient.trackConsent(
+            consentTimestamp: consentData.consentTimestamp,
+            email: consentData.email,
+            userId: consentData.userId,
+            isUserKnown: consentData.isUserKnown
+        ).onSuccess { _ in
+            ITBInfo("Pending consent tracked successfully")
+        }.onError { error in
+            ITBError("Failed to track pending consent: \(error)")
```
joaodordio, line 139: Like we have discussed, it would be good to have a retry mechanism here that tries to track the consent a second time and then fails definitely.

swift-sdk/Internal/UnknownUserManager.swift
```diff
@@ -87,6 +97,60 @@ public class UnknownUserManager: UnknownUserManagerProtocol {
...
+        }
+
+        // Clear the pending consent data
+        pendingConsentData = nil
```
joaodordio, line 143: Would be good to call the clearPendingConsent() method here instead.

joaodordio (changes requested):

PR #940, opened 2025-08-20
🧪 Updates the script to use the latest stable version of iOS available

joaodordio:
#### Changes
- **GitHub Workflow**: 
    - Updated `.github/workflows/build-and-test.yml` to dynamically detect the latest stable iOS 18.x version instead of hardcoding `18.2`

- **Endpoint Tests**: 
    - Updated `tests/endpoint-tests/scripts/run_test.sh` to use the same dynamic detection instead of hardcoding `18.1`


Ayyanchira (approved):
Looking great! I might refer this for codecov in Android

PR #968, opened 2025-11-05
SDK-124 JWT mechanism not engaged via InAppManager

joaodordio:
## 🔹 Jira Ticket(s)

* [SDK-124](https://iterable.atlassian.net/browse/SDK-124)

## ✏️ Description

On `InAppHelper` `getInAppMessagesFromServer` func we were calling the `apiClient.getInAppMessages(NSNumber(value: number))` directly.

This meant that any JWT 401 errors were not being handled with refresh retries, like they are for example when routed through `RequestProcessorUtil`.

That's the main change, everything else is just boiler plate changes to support injecting the AuthManager that's required by the `RequestProcessorUtil`

[SDK-124]: https://iterable.atlassian.net/browse/SDK-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit-tests/InAppHelperTests.swift
+++ b/tests/unit-tests/InAppHelperTests.swift
@@ -108,9 +108,6 @@ class InAppHelperTests: XCTestCase {
         let apiClient = RetryingApiClient()
         apiClient.authManager = authManager
         
-        
-        let initialFailedAuthCount = authManager.failedAuthCount
-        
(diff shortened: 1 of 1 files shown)


sumeruchat:
LGTM! and great work! Clean solution and m glad we added some tests too :)

PR #994, opened 2026-01-13
[SDK-305] Fix build errors in SDK 6.6.4+ with Carthage

sumeruchat:
## Problem

Carthage builds were failing with `SwiftVerifyEmittedModuleInterface` errors when building the SDK with Xcode 16 / Swift 6.2. The errors looked like:

```
error: no type named 'SendRequestValue' in module 'IterableSDK'
error: no type named 'SendRequestError' in module 'IterableSDK'
```

This was blocking Superhuman (high-value customer, red health status) from upgrading to 6.6.4+.

## Root Cause

When Carthage builds frameworks, it sets `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`, which tells Swift to generate `.swiftinterface` files. These interface files can only reference public types.

The issue was that `SendRequestValue` and `SendRequestError` were declared as internal types, but they were being used in return types of protocol methods throughout the SDK. This created a mismatch where the module interface generator couldn't properly expose these types.

## Solution

Made the following types public in `swift-sdk/Internal/RequestSender.swift`:
- `SendRequestValue` (typealias)
- `SendRequestError` (struct, including all properties and methods)

This is the correct fix because these types were already effectively part of the public API through their use in protocol method signatures - we're just making it explicit now.

## Verification

Tested the fix in multiple ways:

1. **Direct xcodebuild test**: Built the SDK target with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` on both iOS device and simulator SDKs - both succeeded.

2. **Actual Carthage build**: Created a test Cartfile pointing to the local repo and ran `carthage update --use-xcframeworks --platform iOS`. Build completed successfully and generated valid XCFrameworks for both `IterableSDK` and `IterableAppExtensions`.

3. **Framework validation**: Verified the built frameworks contain valid Mach-O binaries and Swift module files.

The fix is minimal (only changes access modifiers) and doesn't alter any functionality.

## Related

Fixes [SDK-305](https://iterable.atlassian.net/browse/SDK-305)
Zendesk Ticket: #253251

[SDK-305]: https://iterable.atlassian.net/browse/SDK-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joaodordio (approved):
LGTM

PR #996, opened 2026-01-14
SDK-31 Fix full-position IAM safe area handling

sumeruchat:
## What
Fixes the safe area gap (behind the status bar / Dynamic Island / home indicator) for full-position in-app messages. The background now extends edge-to-edge, matching the HTML content's background.

## Approach
Uses standard `.overFullScreen` modal presentation (already set in `InAppDisplayer`) for all message types. For full-position IAMs, two things ensure the webview content fills the safe area:

1. **`viewport-fit=cover` injection** — The SDK injects `viewport-fit=cover` into the HTML's viewport meta tag before loading. This tells WkWebView the content wants to extend into the safe area. Handles existing viewport tags (appends to content), missing viewport tags (inserts into `<head>`), and HTML without `<head>` (prepends). Supports both single and double quotes in HTML attributes.

2. **`contentInsetAdjustmentBehavior = .never`** — Prevents the WKWebView's scroll view from automatically insetting content away from the safe area, which would override the viewport-fit directive.

3. **Background color fallback** — Sets the VC's `view.backgroundColor` to the message's background color (when available) for full-position IAMs, so any remaining safe area gaps are filled with the correct color during load/animation.

4. **Animation initial bgColor** — For full-position messages, the animation starts with the message's background color instead of `.clear`, preventing a transparent-to-colored flash during fade-in.

## Why this doesn't break non-full-position messages
All changes are gated on `location == .full` or `input.location == .full`:

- **`injectViewportFitCover`**: Only called when `location == .full`. Non-full HTML is loaded unmodified.
- **`contentInsetAdjustmentBehavior`**: Only set to `.never` when `location == .full`. Non-full webviews keep the default behavior.
- **`view.backgroundColor` in `loadView()`**: Only changed for `.full` with a non-nil `backgroundColor`. Non-full messages use the same `initialViewBackgroundColor(isModal:)` as before.
- **`calculateAnimationDetail`**: `initialBgColor` only differs from `.clear` when `input.location == .full`. The existing animation logic for top/center/bottom is completely unchanged.

## Changes
- **`IterableHtmlMessageViewController.swift`** — Added `injectViewportFitCover(html:)` for viewport-fit injection (supports both single and double quotes), `contentInsetAdjustmentBehavior = .never` for full-position, `view.backgroundColor` set to message bgColor for full-position.
- **`InAppCalculations.swift`** — Fixed animation initial bgColor for `.full` to use message's background color instead of `.clear`, preventing transparent flash during fade-in.
- **Test app** — Added POC button and full-screen test view controller for manual verification.

## Testing
**Unit tests:** All 97 tests pass (InAppTests, InAppPresenterTests, InAppParsingTests, IterableHtmlMessageViewControllerTests).

**Manual testing:**
1. Build and run the integration test app
2. Trigger a full-position IAM — verify the background extends edge-to-edge behind the status bar / Dynamic Island
3. Test the "Local Full-Screen IAM (POC)" button — verify identical edge-to-edge behavior
4. Test non-full-position messages (top, center, bottom) — verify no regression in positioning or animation

### DIFF
--- a/tests/unit-tests/IterableHtmlMessageViewControllerTests.swift
+++ b/tests/unit-tests/IterableHtmlMessageViewControllerTests.swift
@@ -214,6 +214,183 @@ class IterableHtmlMessageViewControllerTests: XCTestCase {
         viewController.webView(webView, decidePolicyFor: testAction, decisionHandler: testAction.decisionHandler)
         wait(for: [expectation1], timeout: testExpectationTimeout)
     }
+
+    // MARK: - injectViewportFitCover Tests
+
+    func testViewportFitCover_alreadyPresent_returnsUnchanged() {
+        let html = #"<html><head><meta name="viewport" content="width=device-width, viewport-fit=cover"></head><body></body></html>"#
+        let result = IterableHtmlMessageViewController.injectViewportFitCover(html: html)
+        XCTAssertEqual(result, html)
+    }
+
+    func testViewportFitCover_alreadyPresent_caseInsensitive() {
+        let html = #"<html><head><meta name="viewport" content="width=device-width, Viewport-Fit=Cover"></head><body></body></html>"#
+        let result = IterableHtmlMessageViewController.injectViewportFitCover(html: html)
+        XCTAssertEqual(result, html)
+    }
+
+    func testViewportFitCover_nameFirstViewportMeta_appendsToContent() {
+        let html = #"<html><head><meta name="viewport" content="width=device-width, initial-scale=1.0"></head><body></body></html>"#
+        let result = IterableHtmlMessageViewController.injectViewportFitCover(html: html)
+        XCTAssertTrue(result.contains("width=device-width, initial-scale=1.0, viewport-fit=cover"))
+        // Ensure the rest of the HTML is preserved
+        XCTAssertTrue(result.contains("<body></body></html>"))
+    }
+
+    func testViewportFitCover_contentFirstViewportMeta_prependsToContent() {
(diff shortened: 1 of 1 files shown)


swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -221,6 +235,54 @@ class IterableHtmlMessageViewController: UIViewController {
...
+    /// If a viewport meta tag exists, appends `viewport-fit=cover` to it.
+    /// If no viewport meta tag exists, inserts one into `<head>`.
+    static func injectViewportFitCover(html: String) -> String {
+        // Already has viewport-fit=cover — no change needed
+        if html.range(of: "viewport-fit=cover", options: .caseInsensitive) != nil {
+            return html
+        }
+
+        // Has a viewport meta tag — append viewport-fit=cover to its content
+        if let range = html.range(of: #"(<meta\s+name\s*=\s*"viewport"\s+content\s*=\s*")"#,
```
joaodordio: This regex is using double quotes, htmk with `content='...'` won't work. 

Do we know for sure that the HTML we get on the in-app uses `""` and never `'...'`

swift-sdk/Internal/in-app/InAppPresenter.swift
```diff
@@ -6,79 +6,79 @@ import UIKit
 
 class InAppPresenter {
     static var isPresenting = false
-
```
joaodordio: Nit pick: This file was probably previously modified but now just has whitespaces changes.

PR #997, opened 2026-01-23
SDK-294 Isolate keychain storage per bundle identifier

sumeruchat:
## What
Fixes keychain data collisions when multiple apps share the same keychain access group (e.g., via App Groups). The SDK now uses a bundle-identifier-scoped service name to isolate each app's keychain data.

## Changes
- **KeychainWrapper**: Added `isolatedServiceName()` that appends bundle ID to service name, and `legacyServiceName` for migration compatibility
- **IterableKeychain**: Added `migrateFromLegacy()` to move data from shared storage to isolated storage without overwriting existing data
- **LocalStorage**: Updated to create keychain with both isolated and legacy wrappers, triggers migration during `upgrade()`

## Impact
- **Breaking changes**: None - migration is automatic and transparent
- **Dependencies**: None
- **Performance**: One-time migration during app upgrade, negligible impact

## Testing
**How to test:**
1. Install an older SDK version in two apps sharing keychain access group
2. Set user identity in both apps (they will collide and share data - the bug)
3. Upgrade to this SDK version
4. Verify each app retains its own identity and data is isolated
5. Verify new installs use isolated storage from the start

**Unit tests added:**
- `KeychainWrapperTests`: Tests for isolated service name generation and keychain isolation
- `KeychainMigrationTests`: Tests for migration scenarios (full migration, partial, no overwrite, idempotent)

### DIFF
--- a/swift-sdk/Internal/Utilities/LocalStorage.swift
+++ b/swift-sdk/Internal/Utilities/LocalStorage.swift
@@ -160,9 +160,6 @@ struct LocalStorage: LocalStorageProtocol {
 
         /// moves `email`, `userId`, and `authToken` from `UserDefaults` to `IterableKeychain`
         moveAuthDataFromUserDefaultsToKeychain()
-
-        /// migrates keychain data from legacy service name to isolated service name
-        migrateKeychainToIsolatedStorage()
     }
 
     func migrateKeychainToIsolatedStorage() {


sumeruchat:
@joaodordio Good catch. You're right. Migration only ran on SDK upgrade, so the uninstall/reinstall case would leave legacy keychain data orphaned.

Fixed it. Keychain migration now runs unconditionally on every initialization, outside the version check. Handles the edge case you described.

swift-sdk/Internal/Utilities/LocalStorage.swift
```diff
@@ -150,9 +157,16 @@ struct LocalStorage: LocalStorageProtocol {
     
     func upgrade() {
         ITBInfo()
-        
+
         /// moves `email`, `userId`, and `authToken` from `UserDefaults` to `IterableKeychain`
         moveAuthDataFromUserDefaultsToKeychain()
+
+        /// migrates keychain data from legacy service name to isolated service name
+        migrateKeychainToIsolatedStorage()
```
joaodordio: Do we need to call `migrateKeychainToIsolatedStorage()` here again?

I only see the  `LocalStorage.upgrade()` being called on `InternalIterableAPI.performUpgrade()` which already calls `migrateKeychainToIsolatedStorage()` right?

So the second call inside `upgrade()` here is redundant.

PR #1004, opened 2026-02-06
SDK-341 Retain and pause task runner on 401 JWT

sumeruchat:
## What
When `autoRetry` is enabled and an offline-queued task gets a 401 JWT error, the task is now retained in the DB and the task runner pauses — instead of deleting the task permanently. Processing resumes when a new auth token arrives.

## Changes
- `IterableAPICallTaskProcessor` conditionally returns `.failureWithRetry` on 401 JWT when `autoRetry=true` (previously always `.failureWithNoRetry`)
- `IterableTaskRunner` gains `authPaused` state (separate from network `paused`) that halts processing on JWT 401
- Runner subscribes to `.iterableAuthTokenRefreshed` notification to resume after auth is restored
- `AuthManager` posts `.iterableAuthTokenRefreshed` when a new token is received
- Fixed `scheduleNext()` to set `running=false` before the paused guard (prevents "Already running" race condition on resume)
- Wired `autoRetry` flag from `localStorage` through `DependencyContainer` → `TaskRunner` → `TaskProcessor`

## Impact
- **Breaking changes**: None — all gated behind `autoRetry` flag (defaults to `false`)
- **Dependencies**: Depends on SDK-339 (included in this branch)
- **Performance**: No impact — paused runner does no work until notified

## Testing
**How to test:**
1. Run offline-events tests: `xcodebuild test -scheme swift-sdk -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -only-testing:offline-events-tests`
2. All 61 tests pass including 3 new tests:
   - `testRetainTaskOn401WhenAutoRetryEnabled` — task kept in DB, runner paused
   - `testDeleteTaskOn401WhenAutoRetryDisabled` — task deleted (legacy behavior preserved)
   - `testResumeAfterAuthTokenRefreshed` — runner resumes and processes retained task after auth refresh

### DIFF
--- a/.github/workflows/bcit-integration-test-deep-linking.yml
+++ b/.github/workflows/bcit-integration-test-deep-linking.yml
@@ -80,7 +80,9 @@ jobs:
         ./scripts/setup-local-environment.sh \
           "${{ secrets.BCIT_TEST_PROJECT_ID }}" \
           "${{ secrets.BCIT_ITERABLE_SERVER_KEY }}" \
-          "${{ secrets.BCIT_ITERABLE_API_KEY }}"
+          "${{ secrets.BCIT_ITERABLE_API_KEY }}" \
+          "${{ secrets.BCIT_JWT_API_KEY }}" \
+          "${{ secrets.BCIT_JWT_SECRET }}"
 
     - name: Validate Setup
       working-directory: tests/business-critical-integration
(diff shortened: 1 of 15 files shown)


swift-sdk/Internal/IterableTaskRunner.swift
```diff
@@ -299,13 +322,15 @@ class IterableTaskRunner: NSObject {
...
     private var paused = false
+    private var authPaused = false
     private let networkSession: NetworkSessionProtocol
     private let healthMonitor: HealthMonitor
     private let notificationCenter: NotificationCenterProtocol
     private let timeInterval: TimeInterval
     private let dateProvider: DateProviderProtocol
     private let connectivityManager: NetworkConnectivityManager
     private var running = false
+    private let autoRetry: Bool
```
joaodordio: With this being a `let` on the runner, when the `checkRemoteConfiguration()` is called this value is never updated. Only on the following startup will this be correctly set. Should we have this be updated the same way we do update the locaStorage and the requestHandler on the onSuccess of the getRemoteConfig()?

PR #1057, opened 2026-04-15
SDK-382 BCIT JWT Auth Retry: add real Iterable backend mode

sumeruchat:
## What
The BCIT tester's JWT Auth Retry screen used to fabricate every response locally, so the SDK's new JWT retry logic was never validated against real Iterable traffic. Mirror the Android network-tester so Normal/401 proxy through to `api.iterable.com` while 500/ConnErr stay locally synthesized.

## Changes
- `MockAPIServerURLProtocol.startLoading` dispatches on response mode: Normal and 401 proxy through `URLSession.shared` (handled-marker avoids URLProtocol recursion); 500/ConnErr keep the existing local synthesis.
- For mode 401 the forwarded request's `Authorization` is rewritten to `Bearer <locally-forged expired JWT>` using the same email source (`IterableAPI.email ?? AppDelegate.currentTestEmail`) as `MockAuthDelegate`, so the backend returns real `InvalidJwtPayload` rather than a user-mismatch error.
- `stopLoading` cancels the forwarded task; completion swallows `NSURLErrorCancelled` to respect the URLProtocol contract.
- `MockAPIServer` exposes `jwtSecret`, drops unused success/jwt401 local helpers, and narrows `mockResponse(for:)` to 500/ConnErr.
- Small UI hint next to the response radios: "→ real backend" vs "→ local mock".
- README gains a JWT Auth Retry response-mode table and test-config.json notes.

## Impact
- **Breaking changes**: None — SDK and prod app code untouched; changes are scoped to the BCIT tester app.
- **Dependencies**: None.
- **Performance**: Proxied round-trips introduce real network latency (that is the point).
- **Data**: `Normal` and `401` now send real traffic to whichever Iterable project `test-config.json` points at. Current `projectId` is 28411 — keep using a dedicated test project so prod dashboards stay clean.

## Testing
**How to test:**
1. `tests/business-critical-integration/scripts/build-and-run.sh`
2. Open JWT Auth Retry → enter a test email → **Login**.
3. Mode = **Normal**, tap **Fire All** → network log should show real `200`s from `api.iterable.com` (not the old canned `{"successCount":1}`).
4. Mode = **401**, tap **Fire All** → expect real `401` with `{"code":"InvalidJwtPayload", ...}` from Iterable; confirm SDK pause + refresh + queue flush when flipped back to Normal.
5. Mode = **500** / **Conn Err** → unchanged local behavior.

**Edge cases:**
- Log out (email cleared): proxy in Normal still runs, but the SDK won't attach an auth token — you'll see real 401s from the backend; that's expected.
- Rapid mode flipping during an in-flight request: `stopLoading` should cancel the forwarded task silently (no noisy error).

**Known unverified items:** I have not yet observed a round-trip myself, so POST body forwarding through `URLRequest.httpBody` over URLProtocol is untested end-to-end (known iOS quirk when the SDK uses a body stream). Worth watching the first Normal-mode Fire All for empty-body 400s.

### DIFF
--- a/tests/business-critical-integration/integration-test-app/IterableSDK-Integration-Tester/App/OfflineRetry/MockAPIServer.swift
+++ b/tests/business-critical-integration/integration-test-app/IterableSDK-Integration-Tester/App/OfflineRetry/MockAPIServer.swift
@@ -4,6 +4,14 @@ final class MockAPIServer {
     static let shared = MockAPIServer()
 
     // MARK: - API Response Mode (controlled by JWT Auth Retry panel)
+    //
+    // Matches Android MockServer.ResponseMode + the UI-only CONN_ERROR:
+    // - .normal          → proxy to real Iterable API (see MockAPIServerURLProtocol)
+    // - .jwt401          → proxy to real Iterable API with Authorization swapped
+    //                      to an expired JWT; real backend returns real 401
+    // - .server500       → local synthesized 500 (can't force 500 from prod)
+    // - .connectionError → local synthesized connection error (iOS equivalent of
+    //                      Android's DEAD_PORT trick)
 
     enum APIResponseMode: String, CaseIterable {
         case normal = "Normal"
@@ -13,6 +21,10 @@ final class MockAPIServer {
     }
 
     var apiResponseMode: APIResponseMode = .normal
+
+    /// JWT signing secret used by MockAPIServerURLProtocol when forging the
+    /// expired token for `.jwt401`. Populated by AppDelegate during SDK re-init.
+    var jwtSecret: String?
 
     // MARK: - State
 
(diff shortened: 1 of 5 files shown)


franco-zalamena-iterable:
This is great, but maybe we can simplify it.

The core goal is validating JWT retry against real 401s, and the .jwt401 proxy path does exactly that. But proxying .normal mode adds complexity we don't really need. On android we do proxy on normal mode but almost because it is more complex to not do it on normal mode than it is to just always proxy it. On iOS that does not seem to be the case, we are having a lot of logic just to proxy to normal mode where we don't really want to change anything.

My suggestion is to just to call the real api without interception on normal mode and keep proxy for 401 and mock for 500 and connection error

franco-zalamena-iterable (changes requested):
We can possibly simplify this pr, check above comment

sumeruchat:
Good call @franco-zalamena-iterable — you're right, the iOS equivalent of Android's always-proxy doesn't buy us anything since URLProtocol can just opt out per request.

Pushed `8a11161e`:
- `canInit` returns `false` when mode is `.normal` → the SDK's URLSession hits `api.iterable.com` directly, tester completely out of the way.
- `startLoading` trimmed to only `.jwt401` (proxy + expired JWT) and `.server500` / `.connectionError` (local synth).
- Also kills the POST `httpBody`-through-URLProtocol concern from the original PR description, since we only forward for the 401 path now.
- UI hint updated: "real backend (direct)" / "real backend (expired JWT)" / "local mock". README table updated to match.

tests/business-critical-integration/integration-test-app/IterableSDK-Integration-Tester/App/OfflineRetry/MockAPIServer.swift
```diff
@@ -4,6 +4,14 @@ final class MockAPIServer {
     static let shared = MockAPIServer()
 
     // MARK: - API Response Mode (controlled by JWT Auth Retry panel)
+    //
+    // Matches Android MockServer.ResponseMode + the UI-only CONN_ERROR:
+    // - .normal          → proxy to real Iterable API (see MockAPIServerURLProtocol)
```
joaodordio: This is no longer true.

`MockAPIServerURLProtocol.canInit` returns `false` for `.normal`. It opts out, no proxy involved. 

The `MockAPIServerURLProtocol` header comment (lines 4-5) correctly says "NOT intercepted."

joaodordio (changes requested):

tests/business-critical-integration/integration-test-app/IterableSDK-Integration-Tester/App/OfflineRetry/MockAPIServer.swift
sumeruchat: Good catch — fixed in `9bce3f09`. The `APIResponseMode` docblock now matches the (correct) header comment in `MockAPIServerURLProtocol.swift`.

franco-zalamena-iterable (approved):

joaodordio (approved):

PR #1058, opened 2026-04-16
SDK-392 Resolve Pending on callback drop

joaodordio:
## 🔹 Jira Ticket(s)

* [SDK-392](https://iterable.atlassian.net/browse/SDK-392)

## Root Cause

When `sendRequest()` receives a 401 JWT error, it calls `authManager.scheduleAuthTokenRefreshTimer(successCallback: { attemptSend() })`. 
The retry chain terminates when `AuthManager.requestNewAuthToken()` detects `retryCount >= maxRetry` and returns early, but it does so **without invoking the callback**. The `attemptSend` closure is orphaned, the `Fulfill` is never resolved, and the entire upstream Pending chain hangs.

The same silent-drop pattern exists on several other paths, all of which can leave upstream awaiters stuck forever:
- `scheduleAuthTokenRefreshTimer` when `shouldSkipTokenRefresh` is true (auth paused) or when the user has logged out (no email/userId).
- `requestNewAuthToken` early returns on `shouldPauseRetry`, `pendingAuth` (in‑flight refresh), and `hasFailedAuth`.
- `logoutUser` clearing pending callbacks instead of resolving them.

## Design: Two-callback pattern

Instead of having callers interpret `nil` tokens (the previous approach), the auth manager now accepts two separate callbacks that express the two possible outcomes:
- `successCallback` — called only when a valid token is obtained.
- `onRetryExhausted` — called when retries are permanently exhausted, auth is paused, or the user has logged out while requests were in flight.

The auth manager decides which callback to invoke internally. Callers provide clear intent:
```swift
authManager.scheduleAuthTokenRefreshTimer(
    interval: retryInterval,
    isScheduledRefresh: false,
    successCallback: { _ in attemptSend() },
    onRetryExhausted: { reportFailure(result: result, error: error, ...) }
)
```

[SDK-392]: https://iterable.atlassian.net/browse/SDK-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (approved):
LGTM — the two-callback pattern is clean and the test coverage is thorough. A few minor suggestions:

1. **`clearPendingCallbacks()` is now dead code.** Both call sites were replaced with `invokePendingExhaustionCallbacks()`. The method is redefined as `_ = drainCallbacks()` but never called — can be removed.

2. **`onAuthTokenReceived` flow is dense.** The three-step resolution order (drain piggybacked → queue future refresh → fire direct caller) is correct but hard to follow. A brief doc comment on the method explaining this sequence would help future readers.

3. **`InternalIterableAPI.swift` login path passes `onRetryExhausted: nil`.** If token acquisition fails during login, it silently drops. Might be worth a follow-up to decide if login should surface that failure.

jena-chakour:
Restored for external branch testing

PR #1059, opened 2026-04-24
SDK-297 Route disableDevice through OfflineManager

sumeruchat:
## What
`disableDeviceForCurrentUser` / `disableDeviceForAllUsers` were effectively online-only: if the network was down, the call was dropped. This PR routes both through the offline queue so the SDK retries them automatically when connectivity returns. React Native inherits this by bumping the iOS SDK — no RN work needed.

While doing that, the branch also preserves the original caller identity when request creation / processor selection is deferred, keeps queued `disableDevice` work across logout, and preserves the existing JWT auth-pause behavior for this request class.

## Changes
- `RequestProcessorProtocol` / `RequestHandler` — route both `disableDevice...` calls through the request-processor path. `disableDeviceForCurrentUser` now snapshots the current user identity before deferred processor selection.
- `OfflineRequestProcessor` — add offline implementations for both disable-device methods using the existing queue / replay pipeline, and use the literal `disableDevice` request identifier so successful replays do not reset JWT retry state.
- `OnlineRequestProcessor` / `ApiClient` / `RequestCreator` — thread a `UserIdentitySnapshot` through disable-device request creation so the online fallback (`HealthMonitor.canSchedule() == false`) still targets the user who was current at call time.
- `RequestHandler.handleLogout` / `IterableTaskScheduler` — preserve queued `/disableDevice` tasks across logout while still purging other queued work.
- Tests — keep the existing disable-device request coverage and add regression tests for identity capture, online fallback, auth-retry preservation, and selective logout purge.

## Impact
- **Breaking changes**: None. Public API signatures are unchanged.
- **Dependencies**: None.
- **Performance**: Neutral to negligible. The new work is a small identity snapshot plus a selective queue purge on logout.
- **Data**: Queued `disableDevice` tasks are now preserved across logout so they can replay for the user who just logged out. Other queued tasks still follow the existing purge behavior.

## Testing
**Automated:**
- The existing `testDisableUserforCurrentUser` and `testDisableUserforAllUsers` in `tests/offline-events-tests/RequestHandlerTests.swift` still exercise online-success, online-failure, offline-success, and offline-failure.
- Added regressions:
  - `testDisableDeviceForCurrentUserCapturesIdentityAtCallTime`
  - `testOnlineDisableDeviceForCurrentUserHonorsIdentitySnapshot`
  - `testDisableDeviceForCurrentUserSuccessDoesNotResetAuthRetries`
  - `testDisableDeviceForCurrentUserWithNoIdentityFailsLoudly`
  - `testOfflineDisableDeviceForAllUsersReplaysWithExpectedBody`
  - `testDeleteAllTasksPreservingTasksWithNameKeepsMatchingTasks`

**How to verify locally:**
1. Run the focused offline-events coverage for `RequestHandlerTests` and `TaskSchedulerTests`.
2. Verify `disableDeviceForCurrentUser` while offline, then mutate user state (logout / switch user) before replay; the request should still target the original user.
3. Verify logout preserves queued `/disableDevice` work but purges other queued tasks.
4. Full suite: `./agent_test.sh` — no regressions.

## Manual verification (BCIT tester on real device, JWT auth)

Verified end-to-end on a physical device using the `OfflineRetry` tab in the BCIT integration tester (`tests/business-critical-integration/integration-test-app`). A temporary "Disable Push" button was added to the FreeStyle tab to fire `IterableAPI.disableDeviceForCurrentUser` directly, since BCIT runs with `autoPushRegistration = false` so the Logout button does not auto-fire it.

Steps and observed results:

| # | Action | Expected | Observed |
|---|--------|----------|----------|
| 1 | Login with test email | Auth status reflects logged-in user | ✅ |
| 2 | Register Push | `Device registration succeeded` in Xcode console; SDK has hexToken | ✅ |
| 3 | Go Offline (MockAPIServer → connectionError) | Subsequent API calls fail and route through offline queue | ✅ |
| 4 | Tap Disable Push | `IterableAPI.disableDeviceForCurrentUser` enqueues an offline task | ✅ |
| 5 | Show DB | DB contains a `users/disableDevice` task with the test user's identity in the body | ✅ enqueue confirmed |
| 6 | Tap Logout (top of screen) | Other queued tasks purged; `users/disableDevice` task survives | ✅ **SDK-297 preservation confirmed** |
| 7 | Go Online | TaskRunner resumes, picks up the preserved task | ✅ |
| 8 | Wait ~10s, Show DB | Task replays against the real backend, task is removed from DB | ✅ replay confirmed (~10s) |

The replay landing successfully against `api.iterable.com` proves the body included the original user's identity (the snapshot captured at call time), since the server accepted the disable for that user.


swift-sdk/Internal/api-client/Request/RequestHandler.swift
```diff
@@ -56,18 +58,29 @@ class RequestHandler: RequestHandlerProtocol {
...
                                      onFailure: OnFailureHandler?) -> Pending<SendRequestValue, SendRequestError> {
-        onlineProcessor.disableDeviceForCurrentUser(hexToken: hexToken,
-                                                    withOnSuccess: onSuccess,
-                                                    onFailure: onFailure)
+        // Snapshot identity synchronously so that when `sendUsingRequestProcessor` defers
+        // request construction (waiting on `HealthMonitor.canSchedule()`), the replayed
+        // request still targets the user who was current at call time — not whoever the
+        // live `auth` points to after `logoutPreviousUser()`, `setEmail`, or `setUserId`
+        // has mutated state.
+        let identitySnapshot = UserIdentitySnapshot(auth: authProvider?.auth)
```
joaodordio: One thing that I think is worth flagging here. 

The snapshot pattern correctly handles the documented race (logoutPreviousUser nulling _email/_userId synchronously after the call), but it leaves a fallback hole. 

If `UserIdentitySnapshot(auth: authProvider?.auth)` returns nil for any reason, `RequestCreator.createDisableDeviceRequest` falls back to `setCurrentUser(inDict:)`, which will read the live auth at request-build time. Since `chooseRequestProcessor()` is async, the live auth at build time can already be frim a different user than at call time, which is exactly the type of bug this ticket is fixing.

In reality the public API guards on `isEitherUserIdOrEmailSet()` so the snapshot should be non-nil, but _email/_userId aren't synchronized and there's no structural guarantee that the snapshot stays the only source of truth. Can we add an explicit guard here?

```suggestion
        let identitySnapshot = UserIdentitySnapshot(auth: authProvider?.auth)
        guard identitySnapshot != nil else {
            onFailure?("disableDeviceForCurrentUser called with no current user", nil)
            return SendRequestError.createErroredFuture(reason: "no identity")
        }
         
```

tests/offline-events-tests/TaskSchedulerTests.swift
```diff
@@ -70,6 +70,52 @@ class TaskSchedulerTests: XCTestCase {
...
+                                          networkSession: MockNetworkSession())
+        let scheduler = try createTaskScheduler(notificationCenter: MockNotificationCenter(),
+                                                healthMonitor: healthMonitor)
+
+        scheduler.deleteAllTasks(preservingTasksWithName: preservedName)
+
+        // The scheduler dispatches deletion onto a background CoreData context via
+        // `perform`. Give it a beat to settle, then verify from the main context.
+        let purgeSettled = expectation(description: #function)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { purgeSettled.fulfill() }
```
joaodordio: Small flake-risk callout: this leans on a fixed 1.0s sleep with a 2.0s ceiling to wait for the background perform to settle. On a slow CI runner that's tight. Could we either subscribe to a notification when deletion completes, or poll findAllTasks().count == 1 with a short interval? Same intent, no fixed wait.

tests/offline-events-tests/RequestHandlerTests.swift
```diff
@@ -125,21 +125,162 @@ class RequestHandlerTests: XCTestCase {
...
-        
+
         try handleRequestWithSuccessAndFailure(requestGenerator: requestGenerator,
                                                 path: Const.Path.disableDevice,
                                                 bodyDict: bodyDict)
-        
+
         wait(for: [expectations.successExpectation, expectations.failureExpectation], timeout: testExpectationTimeout)
     }
+
```
joaodordio, line 143: The offline implementation for the all-users variant is in place but I don't see a regression test for it. `RequestHandlerTests` only exercises `forCurrentUser` against the offline processor. Could we add a mirror of `testDisableUserforAllUsers` with `selectOffline: true `(or equivalent) so the offline path for this variant is locked in? 

This is mostly defensive, but given how cheap it is to flip a boolean and re-run the existing assertions, feels worth having.

swift-sdk/Internal/api-client/Request/OnlineRequestProcessor.swift
```diff
@@ -36,20 +36,27 @@ struct OnlineRequestProcessor: RequestProcessorProtocol {
     
     @discardableResult
     func disableDeviceForCurrentUser(hexToken: String,
+                                     identitySnapshot: UserIdentitySnapshot? = nil,
```
joaodordio: Tiny consistency nit: `identitySnapshot: UserIdentitySnapshot? = nil` here, but the protocol declares it without a default and the offline implementation requires it. 

Could we drop the default so the protocol stays the single source of truth for the contract? 

Otherwise a future caller can quietly fall back to the `setCurrentUser` live-auth path without thinking about it, which is exactly the regression we just fixed.

joaodordio (changes requested):
Couple of minor comments but the main request is on `RequestHandler.disableDeviceForCurrentUser`.

swift-sdk/Internal/api-client/Request/RequestHandler.swift
sumeruchat: Good catch — switched to a `guard let` so we fail loud (logs + onFailure + errored future) instead of silently falling back to `setCurrentUser(inDict:)` and reading live auth. Reaching the nil branch now means an internal caller bypassed `isEitherUserIdOrEmailSet()`, so refusing the request is the right behavior. Done in ed596283.

tests/offline-events-tests/TaskSchedulerTests.swift
sumeruchat: Switched to `XCTNSPredicateExpectation` polling on the main context — predicate evaluates on the main thread so `mainContext.findAllTasks()` access is safe, and the wait stops as soon as the purge lands instead of always burning the full 1.0s. 5.0s ceiling for slow CI. Done in ed596283.

tests/offline-events-tests/RequestHandlerTests.swift
sumeruchat: Fair — `testDisableUserforAllUsers` already exercises the offline path via `handleRequestWithSuccessAndFailure` (which fans into online-success/-failure + offline-success/-failure), but the coverage isn't obvious from the name. Added `testOfflineDisableDeviceForAllUsersReplaysWithExpectedBody` as a dedicated visibility test. Done in ed596283.

swift-sdk/Internal/api-client/Request/OnlineRequestProcessor.swift
sumeruchat: Done — dropped the `= nil` default. Protocol is now the single source of truth; a future caller can no longer silently fall through to the old live-auth path. ed596283.

joaodordio (approved):
Codecov lowered the patch coverage on the latest commits because the new fail-loud guard branch in `RequestHandler` is not exercised by any test. 

Adding a quick `testDisableDeviceForCurrentUserWithNoIdentityFailsLoudly` (just mock auth provider with `.none`) would close that gap.

Other than that LGTM!

sumeruchat:
Added `testDisableDeviceForCurrentUserWithNoIdentityFailsLoudly` in ee8da033 — wires a mock auth provider with `.none` identity and asserts `onFailure` fires with the documented reason string and `onSuccess` does not. Should close the codecov gap on the new guard branch. Thanks for the review!

Issue #767, opened 2024-05-03
Swift access race in IterableSDK.Pending.onSuccess


yungkienpoh-prestolabs:
Hi, when I run my project with `ThreadSanitizer`, `Xcode` has discovered a Swift access race issue in `IterableSDK.Pending.onSuccess` function. I am using `Iterable` with version `6.5.2`.

It seems to happen when I set the `userId` but you can further check on this.

Did you know about it and currently working on resolving this issue? Here are the screenshots for better understanding:

<img width="1155" alt="Screenshot 2024-05-03 at 12 24 48 PM" src="https://github.com/Iterable/swift-sdk/assets/152136890/d46a05c4-434b-4354-987b-cfcdea91fa3f">

jena-chakour:
Hey @yungkienpoh-prestolabs hope you're well!
Thanks for flagging this access race, it is addressed in [SDK version 6.7.2](https://github.com/Iterable/iterable-swift-sdk/releases/tag/6.7.2); feel free to let us know your test results.

PR #1063, opened 2026-05-11
SDK-471 Fix Swift access race in Pending/Fulfill

sumeruchat:
## What
Thread Sanitizer flagged a Swift access race inside `Pending` / `Fulfill` (reproducible in customer apps, GitHub issue #767, easiest repro via `setUserId`). All mutable state — `successCallbacks`, `errorCallbacks`, and `result` — was accessed with zero synchronization while URLSession delegate queues, persistenceContext, and main thread routinely fan into the same instance.

This change serializes all state mutations behind a private `DispatchQueue` (matching the established `AuthManager.callbackQueue` pattern) and snapshots callbacks before invoking them outside the critical section so re-entrant registrations cannot deadlock. Multi-event semantics of `Fulfill` (used by `MockInAppDisplayer.onShow` and similar in-tree callers) are preserved.

## Changes
- **`swift-sdk/Internal/Pending.swift`** — All reads/writes of `result`, `successCallbacks`, `errorCallbacks` now run inside `stateQueue.sync`. Callbacks fire OUTSIDE the sync block. `Fulfill.resolve` / `reject` route through a new `setResult(_:)` instead of relying on `didSet`. Repeated resolution / rejection still notifies every registered callback (`Fulfill` is still usable as an event signal — no behavior change).
- **`onCompletion` atomicity** — both callbacks are appended inside one sync block; a `.failure` resolve landing between the two appends can no longer lose the error handler.
- **Late registration on resolved Pending** — appends the callback and replays the stored result for the new callback only (previously iterated the full accumulated array, causing O(N²) firing). Late-registered callbacks remain in the list and receive subsequent resolves.
- **`tests/unit-tests/PendingConcurrencyTests.swift`** — 11 new tests covering concurrent register+resolve, `onCompletion` atomicity under race (deterministic via `DispatchGroup`), re-entrant callback registration, flatMap cross-queue resolution, late-register-after-resolve, repeated resolve/reject behaviour, mixed resolve-after-reject, and a 1000-iteration stress run.

## Impact
- **Breaking changes**: None. The latent O(N²) firing on late-register-against-resolved-Pending is corrected — each late-registered callback now fires once per resolution rather than accumulating. No in-tree caller depended on the O(N²) behaviour; all existing tests pass.
- **Dependencies**: None.
- **Performance**: Net win — the O(N²) firing path is gone. Sync overhead is a single uncontended `dispatch_sync` per operation.

## Testing
**Automated (verified locally):**
- `./agent_test.sh PendingTests` — all 9 original tests pass
- `./agent_test.sh PendingConcurrencyTests` — all 11 new tests pass
- `./agent_test.sh InAppTests` — 43/43 pass (covers `MockInAppDisplayer.onShow` multi-event use)
- `./agent_test.sh AuthTests` — 37/37 pass

**How to verify TSan locally:**
1. Open `swift-sdk.xcodeproj` in Xcode
2. Edit `swift-sdk` scheme → Test → Diagnostics → enable Thread Sanitizer
3. Run `PendingConcurrencyTests` — expect zero TSan reports
4. Optional: run a customer-style `setUserId` flow against a sample app with TSan enabled — the original prestolabs warning should be gone

**Edge cases covered:**
- Resolve racing N=50 `onSuccess` registrations from a concurrent queue
- Reject racing N=50 `onError` registrations
- `onCompletion(value:error:)` partial-append race (200 iterations, `DispatchGroup`-synchronized)
- Re-entrant `onSuccess` registration from inside an executing callback
- `flatMap` chain crossing 3 queues (background → main → background)
- Repeated `resolve` / `reject` notifies every registered callback
- Mixed `reject` then `resolve` notifies the matching branch each time
- Late-registered callbacks stay registered and receive subsequent resolves
- `concurrentPerform`-based stress run of 1000 Fulfills

### DIFF
--- a/swift-sdk.xcodeproj/project.pbxproj
+++ b/swift-sdk.xcodeproj/project.pbxproj
@@ -405,6 +405,7 @@
 		ACE34AB5213776CB00691224 /* LocalStorageTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACE34AB4213776CB00691224 /* LocalStorageTests.swift */; };
 		ACED4C01213F50B30055A497 /* LoggingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACED4C00213F50B30055A497 /* LoggingTests.swift */; };
 		ACEDF41F2183C436000B9BFE /* PendingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACEDF41E2183C436000B9BFE /* PendingTests.swift */; };
+		5C0471F221F0F0F0000B7CCC /* PendingConcurrencyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5C0471F121F0F0F0000B7CCC /* PendingConcurrencyTests.swift */; };
 		ACF406252507F90F005FD775 /* NetworkConnectivityManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACF406242507F90F005FD775 /* NetworkConnectivityManagerTests.swift */; };
 		ACF560D620E443BF000AAC23 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACF560D520E443BF000AAC23 /* AppDelegate.swift */; };
 		ACF560DB20E443BF000AAC23 /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = ACF560D920E443BF000AAC23 /* Main.storyboard */; };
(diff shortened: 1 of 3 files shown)


joaodordio (approved):

sumeruchat:
follow-up commit pushed (`34a44ff0`) — purely test hardening, no SDK change. the original `testFlatMapCrossQueueResolution` registered `onCompletion` on a `.background` queue async, which exceeded the 15s wait under GHA runner contention (run 25661208588). registering synchronously removes the scheduling dependency; concurrent registration coverage is preserved by `testConcurrentOnSuccessAndResolve` in the same file. CI rerun on the previous commit also passed (run 26092897516) — confirmed it was a transient flake. thanks for the approval 🙏

PR #1064, opened 2026-05-13
SDK-466 Cache DateFormatter and skip formatting for filtered log levels

sumeruchat:
## Summary
- Cache `IterableLogUtil`'s `DateFormatter` as a `private static let` so it is created once instead of on every log call (fixes the 2.4–3.2s main-thread hang reported by Wolt via Sentry during foreground transitions).
- Add an optional `shouldLog(level:) -> Bool` to `IterableLogDelegate` (default `true` when unimplemented) and short-circuit `IterableLogUtil.log()` before `formatLogMessage()` when the delegate says no — so apps with `DefaultLogDelegate(minLogLevel: .error)` no longer pay format cost for filtered `ITBInfo()`/`ITBDebug()` calls.
- `DefaultLogDelegate` routes its `minLogLevel` filter through `shouldLog`; `NoneLogDelegate` returns `false` for every level (bonus perf win).

Ticket: [SDK-466](https://iterable.atlassian.net/browse/SDK-466)

## Test plan
- [x] `./agent_build.sh` passes
- [x] `./agent_test.sh` passes (unit, notification-extension, and offline-events bundles all 0 failures)
- [x] New `LoggingTests` cover the short-circuit: `SpyDateProvider` confirms `currentDate` is never accessed when `DefaultLogDelegate` filters the level out, and never accessed at all under `NoneLogDelegate`
- [ ] Reviewer: sanity-check the new Obj-C selector name `shouldLog:` on `IterableLogDelegate` — low risk (Iterable-owned protocol, matches existing `log:message:` convention) but worth eyeballing

[SDK-466]: https://iterable.atlassian.net/browse/SDK-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/SDK/IterableConfig.swift
+++ b/swift-sdk/SDK/IterableConfig.swift
@@ -85,6 +85,12 @@ public struct IterableAPIMobileFrameworkInfo: Codable {
     ///     - message: The message to log. The message will include file, method and line of the call.
     @objc(log:message:)
     func log(level: LogLevel, message: String)
+
+    /// Return whether a message at the given level should be formatted and logged.
+    /// If this optional method is not implemented, all levels will be formatted and logged.
+    /// - Parameter level: The logging level
+    @objc(shouldLog:)
+    optional func shouldLog(level: LogLevel) -> Bool
 }
 
 /// The delegate for getting the authentication token
--- a/swift-sdk/Internal/Utilities/IterableLogUtil.swift
+++ b/swift-sdk/Internal/Utilities/IterableLogUtil.swift
@@ -16,6 +16,8 @@ struct IterableLogUtil {
     static var sharedInstance: IterableLogUtil?
     
     func log(level: LogLevel, message: String?, file: String, method: String, line: Int) {
+        guard logDelegate.shouldLog?(level: level) ?? true else { return }
+
         let logMessage = IterableLogUtil.formatLogMessage(message: message, file: file, method: method, line: line, date: dateProvider.currentDate)
         logDelegate.log(level: level, message: logMessage)
     }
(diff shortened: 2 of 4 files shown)


swift-sdk/SDK/IterableConfig.swift
```diff
@@ -85,6 +85,12 @@ public struct IterableAPIMobileFrameworkInfo: Codable {
     ///     - message: The message to log. The message will include file, method and line of the call.
     @objc(log:message:)
     func log(level: LogLevel, message: String)
+
+    /// Return whether a message at the given level should be formatted and logged.
+    /// If this optional method is not implemented, all levels will be formatted and logged.
+    /// - Parameter level: The logging level
+    @objc(shouldLog:)
+    optional func shouldLog(level: LogLevel) -> Bool
```
franco-zalamena-iterable, line 93: isn't it possible to have this returning true by default and if overwritten then it changes instead of having it null?

franco-zalamena-iterable (changes requested):

swift-sdk/SDK/IterableConfig.swift
sumeruchat: good catch — i went back and forth on this one. the constraint: `IterableLogDelegate` is `@objc` (line 81 in `IterableConfig.swift`), and `@objc` protocols can't get a default impl that's visible to Obj-C subclasses. Swift protocol extensions only work for pure-Swift conformers — for Obj-C log delegates the method would become required, which is a breaking change for anyone implementing the protocol from Obj-C today.

`@objc optional` was the way to keep the Swift call site reasonably clean (`?? true`) while leaving existing Obj-C consumers compiling unchanged. happy to revisit if we'd rather drop Obj-C overrideability on this specific method — but i'd lean toward keeping the back-compat over the call-site nit. wdyt?

franco-zalamena-iterable: ok, i didn't know about the objc limitation. I think this is a good middle ground

franco-zalamena-iterable (approved):

Issue #867, opened 2024-11-27
NetworkMonitor memory leak on IterableTaskRunner start


tristan-warner-smith:
`Configuration:`
* Iterable `6.5.7`
* iPhone 16 Pro (18.2 simulator)
* Xcode 16.2 beta 3

SwiftUI project with `UIApplicationDelegate`.
Iterable is configured in `didFinishLaunchingWithOptions`, other related calls in the delegate methods.

`Report:`
Instruments shows `IterableTaskRunner.start() > NetworkConnectivityManager.start() > NetworkMonitor.start()` as the cause of the leak.

![Leak](https://github.com/user-attachments/assets/cf623457-0064-4b3e-9126-8592fb1d4e8e)

jena-chakour:
@tristan-warner-smith thanks for surfacing this!
We have made an internal note to address this in upcoming SDK improvements.

jena-chakour:
@tristan-warner-smith quick followup on this, can your team please verify if the same is seen using the [latest SDK version](https://github.com/Iterable/iterable-swift-sdk/releases/tag/6.5.12)?

jena-chakour:
Hey @tristan-warner-smith hope you're well!
[SDK version 6.7.2](https://github.com/Iterable/iterable-swift-sdk/releases/tag/6.7.2) fixes this memory leak. Thanks again for reaching out about it, feel free to let us know your test results.

PR #1065, opened 2026-05-14
SDK-468 Fix NetworkMonitor memory leak

sumeruchat:
## Summary
- Fixes a customer-reported memory leak (Iterable 6.5.7 / 6.5.12, iOS 17.5 / 18.2, GitHub issue #867) where `IterableTaskRunner.start()` → `NetworkConnectivityManager.start()` → `NetworkMonitor.start()` retained `NWPathMonitor` + the `NetworkMonitor` wrapper for the lifetime of the app. Three compounding causes: `pathUpdateHandler` captured `self` strongly, the stored monitor reference was `weak` (so `stop()` was unreliable), and `start()` was not idempotent across foreground/background cycles.
- `NetworkMonitor.start()` is now idempotent, captures `[weak self]` in `pathUpdateHandler`, and stores `NWPathMonitor` strongly via a small `NetworkPathMonitorProtocol` wrapper injected through a `pathMonitorFactory` closure (default = real impl). `stop()` clears the handler, cancels the monitor, and nils the reference.
- `NetworkConnectivityManager` cleanups: `stop()` clears `statusUpdatedCallback`, `startTimer()` is idempotent, and `updateStatus()`'s connectivity-check success callback uses `[weak self]`.
- Includes a small `agent_test.sh` improvement: resolve the simulator by UDID and boot via `xcrun simctl` so xcodebuild doesn't fail when multiple simulators share a model name across runtimes. **Non-intrusive** — does not touch any `Simulator.app` instance the developer may have open for another project.

Ticket: [SDK-468](https://iterable.atlassian.net/browse/SDK-468) · Closes GitHub issue #867

## Test plan
- [x] `./agent_build.sh` passes
- [x] `./agent_test.sh` passes (588 unit + 12 notification-extension + 90 offline-events = 690 tests, 0 failures)
- [x] New `NetworkMonitorLifecycleTests` (3 cases): dealloc after `start()`/`stop()`, first monitor released on consecutive `start()` calls, no callback fires after `stop()` (also asserts `cancel()` was called)
- [x] New `TaskRunnerTests.testForegroundBackgroundCyclingDoesNotLeakConnectivityLifecycle`: cycles `didEnterBackground` / `willEnterForeground` 3× and asserts task runner, manager, and monitor all deallocate (weak-ref via `WeakBox`)
- [ ] Reviewer: the `NetworkPathMonitorProtocol` wrapper is a new internal type added for testability — please sanity-check that you're happy with the shape
- [ ] Reviewer (optional follow-up): run the original SwiftUI repro under Instruments → Leaks for 60s to confirm zero `NWPathMonitor` / `NetworkMonitor` leaks rooted in `NetworkMonitor.start()`

[SDK-468]: https://iterable.atlassian.net/browse/SDK-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/Network/NetworkMonitor.swift
+++ b/swift-sdk/Internal/Network/NetworkMonitor.swift
@@ -8,6 +8,45 @@ import Foundation
 import Network
 #endif
 
+struct NetworkPathUpdate {
+    let debugDescription: String
+    let status: String
+}
+
+protocol NetworkPathMonitorProtocol: AnyObject {
+    var pathUpdateHandler: ((NetworkPathUpdate) -> Void)? { get set }
+    func start(queue: DispatchQueue)
+    func cancel()
+}
+
+#if canImport(Network)
+private final class NetworkPathMonitor: NetworkPathMonitorProtocol {
+    var pathUpdateHandler: ((NetworkPathUpdate) -> Void)? {
+        didSet {
+            guard let pathUpdateHandler = pathUpdateHandler else {
+                monitor.pathUpdateHandler = nil
+                return
+            }
+
+            monitor.pathUpdateHandler = { path in
+                pathUpdateHandler(NetworkPathUpdate(debugDescription: path.debugDescription,
+                                                    status: String(describing: path.status)))
+            }
+        }
+    }
+
+    func start(queue: DispatchQueue) {
+        monitor.start(queue: queue)
+    }
+
+    func cancel() {
+        monitor.cancel()
+    }
+
+    private let monitor = NWPathMonitor()
+}
+#endif
+
 /// Listens to network interface to detect status change.
 /// It only knows that the status has changed.
 /// It does not know if the network is online or not.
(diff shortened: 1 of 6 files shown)


swift-sdk/Internal/Network/NetworkMonitor.swift
```diff
@@ -43,10 +85,12 @@ class NetworkMonitor: NetworkMonitorProtocol {
     
     func stop() {
         ITBInfo()
+        networkMonitor?.pathUpdateHandler = nil
         networkMonitor?.cancel()
         networkMonitor = nil
     }
```
joaodordio, line 95: Small asymmetry: `NetworkConnectivityManager.stop()` now clears its inner `statusUpdatedCallback`, but `NetworkMonitor.stop()` does not clear its own. The `pathUpdateHandler` is already weak-self so it's not a leak, but for symmetry and to make `stop()` a true reset, consider also nilling `statusUpdatedCallback` here (or add a one-line comment explaining why we keep it set across stop/start cycles).

joaodordio (approved):

swift-sdk/Internal/Network/NetworkMonitor.swift
sumeruchat: thanks @joaodordio — went with the comment route. pushed `64338af4` with a 3-line explainer in `stop()`: `statusUpdatedCallback` is intentionally preserved across stop/start cycles so callers (`NetworkConnectivityManager` etc.) don't have to re-register on every cycle, and the weak-self capture in `pathUpdateHandler` keeps it safe.

PR #1067, opened 2026-05-22
SDK-99 Add async/await variants for disablePush

sumeruchat:
## What
Adds Swift concurrency overloads for the disablePush APIs so callers can `await` them. This was the only remaining piece of SDK-99 after offline-retry (SDK-297) and the success/fail callbacks already shipped.

## Changes
- Add `disableDeviceForCurrentUser() async throws` and `disableDeviceForAllUsers() async throws` on `IterableAPI`, wrapping the existing completion-handler APIs via `withCheckedThrowingContinuation`.
- Throw the public `SendRequestError` on failure (carries the reason/data from `OnFailureHandler`); an `NSLock`-backed guard makes resume exactly-once.
- Preflight throws `"Iterable SDK is not initialized"` instead of hanging the `await` on the existing silent not-initialized path (the completion methods call neither handler when uninitialized).
- `@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)` + `@nonobjc` so existing methods and the Obj-C surface stay untouched (base target remains iOS 10).

## Impact
- **Breaking changes**: None — purely additive overloads; existing parameterless/completion-handler methods and Obj-C selectors are unchanged.
- **Dependencies**: None.
- **Performance**: Negligible — a thin continuation wrapper over the existing call path.

## Testing
**How to test:**
1. Run `./agent_build.sh` (succeeds) and `./agent_test.sh` (unit 605 / notification-extension 12 / offline-events 90, 0 failures).
2. From an iOS 13+ target after `IterableAPI.initialize…`: `try await IterableAPI.disableDeviceForCurrentUser()` → a `disableDevice` request is sent with the token.
3. Force a server failure (e.g. 400) → the call throws `SendRequestError` carrying the response reason/data.
4. Call before initializing the SDK → throws `"Iterable SDK is not initialized"` and does **not** hang.

Jira: https://iterable.atlassian.net/browse/SDK-99

### DIFF
--- a/swift-sdk/SDK/IterableAPI.swift
+++ b/swift-sdk/SDK/IterableAPI.swift
@@ -300,10 +300,28 @@ import UIKit
     public static func disableDeviceForCurrentUser() {
         disableDeviceForCurrentUser(withOnSuccess: nil, onFailure: nil)
     }
+
+    /// Disable this device's token in Iterable, for the current user.
+    @nonobjc
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+    public static func disableDeviceForCurrentUser() async throws {
+        try await disableDeviceAsync { onSuccess, onFailure in
+            disableDeviceForCurrentUser(withOnSuccess: onSuccess, onFailure: onFailure)
+        }
+    }
     
     /// Disable this device's token in Iterable, for all users on this device.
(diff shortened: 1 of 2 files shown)


franco-zalamena-iterable:
Does iOS have changelog entries only on the release or should we add it in this pr?

franco-zalamena-iterable (approved):

PR #1068, opened 2026-05-26
SDK-108 Route registerToken through OfflineRequestProcessor for network-failure retry

sumeruchat:
## What
Brings `registerToken` to feature parity with `disablePush` (shipped under SDK-297): the registration request is now persisted to the offline task queue and replayed when the network returns, instead of being dropped fire-once on transient network failures. Customer-visible symptom this fixes: device never gets registered for push if `registerForRemoteNotifications` is called while the device is offline.

Jira: https://iterable.atlassian.net/browse/SDK-108

## Changes
- **`RequestProcessorProtocol.swift`** — add `register(registerTokenInfo:notificationsEnabled:onSuccess:onFailure:)` to the shared protocol so `sendUsingRequestProcessor` can pick offline vs online at call time. Add `RequestIdentifier.registerToken` (sibling to the existing `disableDevice` identifier).
- **`RequestHandler.swift`** — resolve `notificationStateProvider.isNotificationsEnabled` *first*, then route through `sendUsingRequestProcessor` with the resolved bool. This snapshots `notificationsEnabled` at call time (parallel to SDK-297's `UserIdentitySnapshot`), so a request queued offline and replayed later uses the bool as it was when the caller invoked register — replay does **not** re-query notification state.
- **`OfflineRequestProcessor.swift`** — implements the new `register(...)` using the existing `RequestCreator.createRegisterTokenRequest(...)` + `sendIterableRequest(...)` pattern. Drop-in mirror of `disableDeviceForCurrentUser`.
- **`OnlineRequestProcessor.swift`** — public `register(...)` now takes `notificationsEnabled: Bool` directly (provider resolution moved up to `RequestHandler`); the redundant private helper is folded into the public method; uses `RequestIdentifier.registerToken` consistently.

## Impact
- **Breaking changes**: None. Public `IterableAPI` / `InternalIterableAPI` `register(token:onSuccess:onFailure:)` surface and Obj-C selectors are unchanged.
- **Dependencies**: None.
- **Performance**: Negligible. Same number of network calls in the success path; offline path now persists one row in the existing `IterableTaskScheduler` CoreData queue instead of dropping the request.
- **Snapshot semantics**: explicitly tested — `notificationsEnabled` is frozen at call time. If notification permissions change between schedule and replay, the replayed request still carries the original bool. This matches the SDK-297 precedent and is the desired behavior for queued events.

## Testing
**How to test:**
1. Run `./agent_build.sh` (succeeds) and `./agent_test.sh` (601 unit / 12 notification-extension / 92 offline-events, 0 failures).
2. New `testOfflineRegisterTokenReplaysWithExpectedBody` — confirms an offline-mode register persists and replays with the expected body.
3. New `testOfflineRegisterTokenRetriesAfterNetworkFailureWithSnapshottedNotificationsEnabled` — sends a register, fails the first network attempt, mutates `MockNotificationStateProvider.enabled` from `true → false`, asserts both the failed first send and the replayed second send carry `notificationsEnabled: true` (the snapshotted value) and the task is removed from the queue after the successful replay.
4. Manual sanity: with the app in airplane mode, call `registerForRemoteNotifications`; turn airplane mode off; observe one `register-device-token` request fired with the original notification-enabled bool.

### DIFF
--- a/swift-sdk/Internal/api-client/Request/RequestHandler.swift
+++ b/swift-sdk/Internal/api-client/Request/RequestHandler.swift
@@ -48,10 +48,20 @@ class RequestHandler: RequestHandlerProtocol {
                   notificationStateProvider: NotificationStateProviderProtocol,
                   onSuccess: OnSuccessHandler?,
                   onFailure: OnFailureHandler?) {
+        // Snapshot identity before the async notification-state callback so deferred
+        // request construction still targets the call-time user.
+        guard let identitySnapshot = UserIdentitySnapshot(auth: authProvider?.auth) else {
+            let reason = "register(token:) called without a current user identity"
+            ITBError(reason)
+            onFailure?(reason, nil)
+            return
+        }
+
         notificationStateProvider.isNotificationsEnabled { notificationsEnabled in
             _ = self.sendUsingRequestProcessor { processor in
                 processor.register(registerTokenInfo: registerTokenInfo,
                                    notificationsEnabled: notificationsEnabled,
+                                   identitySnapshot: identitySnapshot,
                                    onSuccess: onSuccess,
                                    onFailure: onFailure)
             }
--- a/swift-sdk/Internal/api-client/Request/RequestHandler.swift
+++ b/swift-sdk/Internal/api-client/Request/RequestHandler.swift
@@ -48,10 +48,14 @@ class RequestHandler: RequestHandlerProtocol {
                   notificationStateProvider: NotificationStateProviderProtocol,
                   onSuccess: OnSuccessHandler?,
                   onFailure: OnFailureHandler?) {
-        onlineProcessor.register(registerTokenInfo: registerTokenInfo,
-                                 notificationStateProvider: notificationStateProvider,
-                                 onSuccess: onSuccess,
-                                 onFailure: onFailure)
+        notificationStateProvider.isNotificationsEnabled { notificationsEnabled in
+            _ = self.sendUsingRequestProcessor { processor in
+                processor.register(registerTokenInfo: registerTokenInfo,
+                                   notificationsEnabled: notificationsEnabled,
+                                   onSuccess: onSuccess,
+                                   onFailure: onFailure)
+            }
+        }
     }
     
     @discardableResult
--- a/swift-sdk/Internal/api-client/Request/RequestHandler.swift
+++ b/swift-sdk/Internal/api-client/Request/RequestHandler.swift
@@ -50,18 +50,24 @@ class RequestHandler: RequestHandlerProtocol {
                   onFailure: OnFailureHandler?) {
         // Snapshot identity before the async notification-state callback so deferred
         // request construction still targets the call-time user.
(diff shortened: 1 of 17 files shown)


swift-sdk/Internal/api-client/Request/RequestHandler.swift
```diff
@@ -48,10 +48,14 @@ class RequestHandler: RequestHandlerProtocol {
                   notificationStateProvider: NotificationStateProviderProtocol,
                   onSuccess: OnSuccessHandler?,
                   onFailure: OnFailureHandler?) {
-        onlineProcessor.register(registerTokenInfo: registerTokenInfo,
-                                 notificationStateProvider: notificationStateProvider,
-                                 onSuccess: onSuccess,
-                                 onFailure: onFailure)
+        notificationStateProvider.isNotificationsEnabled { notificationsEnabled in
+            _ = self.sendUsingRequestProcessor { processor in
```
franco-zalamena-iterable, line 68: On master, register went straight to OnlineRequestProcessor and built the body synchronously after isNotificationsEnabled, so setCurrentUser read auth essentially at call time. After this PR, the build runs inside sendUsingRequestProcessor's deferred closure (after canSchedule() + a queue hop), and createRegisterTokenRequest still reads live authProvider.auth rather than a captured UserIdentitySnapshot like disableDeviceForCurrentUser does. Intentional, or should register mirror the SDK-297 snapshot pattern?

franco-zalamena-iterable (changes requested):

swift-sdk/Internal/api-client/Request/RequestHandler.swift
sumeruchat: Mirrored the snapshot pattern in fb361ff4. `register` now captures `UserIdentitySnapshot` at call time (before the async notification hop) and threads it into `createRegisterTokenRequest`, same as `disableDeviceForCurrentUser`. Tests cover offline-replay and online-fallback keeping the call-time identity. 36/0.

franco-zalamena-iterable: Optional UserIdentitySnapshot? on six signatures + a forked branch in RequestCreator feels heavy for a race that only exists in RequestHandler.register's async hop. The disableDevice symmetry is also partial, those tasks survive handleLogout(), register doesn't. Can we fix it where the gap actually is instead of threading a parameter through every layer below it?

sumeruchat: Simplified in eed546e2. Dropped the cross-layer snapshot threading: capture call-time auth once in `register`, carry it on `RegisterTokenInfo.auth`, and `RequestCreator` resolves it through the existing single body path. `ApiClient`, both processors, and the protocol no longer carry the param. You were right on `handleLogout`, register tasks are purged on logout (only disableDevice is preserved), so the full symmetry was not needed. Kept a small identity guard in `register` because dropping it regresses `onFailure` on the offline no-user path. 36/0.

franco-zalamena-iterable (approved):

PR #1069, opened 2026-05-27
SDK-300 Add OnSuccess/OnFailure handlers to logoutUser

sumeruchat:
## What
Adds a public `IterableAPI.logoutUser(withOnSuccess:onFailure:)` overload so SDK consumers can observe the logout sequence. The handlers signal that the local logout sequence completed and surface the triggered `disableDeviceForCurrentUser` result — they are an observability primitive, not a retry mechanism (the disable already auto-retries offline via SDK-297). Ticket: [SDK-300](https://iterable.atlassian.net/browse/SDK-300).

## Changes
- **Public API** (`IterableAPI.swift`): New `logoutUser(withOnSuccess:onFailure:)` overload. Parameterless `logoutUser()` now delegates with `nil/nil` handlers, preserving Obj-C and source compatibility. Mirrors `disableDeviceForCurrentUser`'s two-overload precedent.
- **Internal** (`InternalIterableAPI.swift`): New `logoutUser(withOnSuccess:onFailure:)`. Not-initialized → `onFailure`. `autoPushRegistration` on → handlers piped through the existing `disableDeviceForCurrentUser(withOnSuccess:onFailure:)` **before** identity cleanup (so the disable request goes out with the still-set userId/email), then local cleanup runs. `autoPushRegistration` off → local cleanup, then `onSuccess(nil)` synchronously.
- **`logoutPreviousUser()` untouched** — internal user-switch paths keep their existing fire-and-forget contract.
- **Tests** (`AuthTests.swift`): 5 new tests covering autoPush-on success, autoPush-on failure (verifies local cleanup STILL happens despite network error), autoPush-off synchronous success with no network, not-initialized failure, and the parameterless overload regression.

## Impact
- **Breaking changes**: None. The existing parameterless `logoutUser()` is unchanged; only adds a new overload.
- **Dependencies**: None.
- **Performance**: No-op — handlers are only invoked at the existing completion points.

## Testing
**How to test:**
1. Run `./agent_test.sh AuthTests` — all 42 tests pass (5 new).
2. Call `IterableAPI.logoutUser(withOnSuccess:onFailure:)` from a sample app with `autoPushRegistration = true` and a logged-in user; verify `onSuccess` fires after the disable network call returns.
3. With `autoPushRegistration = false`, verify `onSuccess` fires synchronously and no network request is made.
4. Without initializing the SDK (or without a logged-in user), verify `onFailure` fires with `Iterable SDK is not initialized`.
5. Confirm the existing parameterless `logoutUser()` still logs out without regression.

[SDK-300]: https://iterable.atlassian.net/browse/SDK-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -248,7 +248,37 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
     }
 
     func logoutUser() {
-        logoutPreviousUser()
+        logoutUser(withOnSuccess: nil, onFailure: nil)
+    }
+
+    func logoutUser(withOnSuccess onSuccess: OnSuccessHandler?,
+                    onFailure: OnFailureHandler?) {
+        ITBInfo()
+
+        guard isSDKInitialized() else {
+            onFailure?("Iterable SDK is not initialized", nil)
+            return
+        }
+
+        if config.autoPushRegistration {
+            disableDeviceForCurrentUser(withOnSuccess: onSuccess, onFailure: onFailure)
+        }
+
+        _email = nil
+        _userId = nil
+
+        storeIdentifierData()
+
+        authManager.logoutUser()
+
+        _ = inAppManager.reset()
+        _ = embeddedManager.reset()
+
+        try? requestHandler.handleLogout()
+
+        if !config.autoPushRegistration {
+            onSuccess?(nil)
+        }
     }
     
     func attemptAndProcessMerge(merge: Bool, replay: Bool, destinationUser: String?, isEmail: Bool, failureHandler: OnFailureHandler? = nil) {
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -879,7 +879,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         // sequence has a single source of truth. The user-switch paths (setEmail/
         // setUserId) pass no handlers: a nil onFailure keeps the not-initialized
         // guard a silent no-op, and a nil onSuccess makes the auto-push-off
-        // completion a no-op — matching this method's previous behavior.
+        // completion a no-op, matching this method's previous behavior.
         logoutUser(withOnSuccess: nil, onFailure: nil)
     }
     
--- a/swift-sdk/Internal/InternalIterableAPI.swift
+++ b/swift-sdk/Internal/InternalIterableAPI.swift
@@ -248,7 +248,7 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
     }
 
     func logoutUser() {
-        logoutUser(withOnSuccess: nil, onFailure: nil)
+        logoutPreviousUser()
     }
 
     func logoutUser(withOnSuccess onSuccess: OnSuccessHandler?,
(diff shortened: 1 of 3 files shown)


swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -248,7 +248,37 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
     }
 
     func logoutUser() {
-        logoutPreviousUser()
```
franco-zalamena-iterable, line 251: Before this PR, logoutUser() delegated to logoutPreviousUser() (line 251 on master), keeping a single source of truth for the eight-step logout cleanup sequence. After this PR, that body lives in two places:
- logoutPreviousUser() (line 877), used by setEmail/setUserId user-switch paths.
- logoutUser(withOnSuccess:onFailure:) (line 254), used by both public entry points.

franco-zalamena-iterable (changes requested):

swift-sdk/Internal/InternalIterableAPI.swift
sumeruchat: @franco-zalamena-iterable Good catch and appreciate the attention to detail. This is fixed in c0193b70. `logoutPreviousUser()` now delegates to `logoutUser(withOnSuccess:onFailure:)` with nil handlers instead of carrying its own copy of the eight-step cleanup, so the sequence has a single source of truth again.

Behavior is unchanged for the setEmail/setUserId user-switch paths: a nil `onFailure` keeps the not-initialized guard a silent no-op, and a nil `onSuccess` makes the auto-push-off completion a no-op — exactly matching the old `logoutPreviousUser()` body.

Verified: build green, AuthTests 42/0, and UserMergeScenariosTests 21/0 (which exercises the user-switch paths through `logoutPreviousUser`).

franco-zalamena-iterable, line 251: don't we have to revert this now? I know it's mostly aesthetic, but with your change it would make sense to revert this change here

franco-zalamena-iterable (approved):
Just one thing to address but non blocker anymore

swift-sdk/Internal/InternalIterableAPI.swift
sumeruchat: Good call — reverted in f7dcf417. `logoutUser()` now calls `logoutPreviousUser()` exactly as on master, and since `logoutPreviousUser()` delegates onward to `logoutUser(withOnSuccess:onFailure:)`, the behavior is unchanged: `logoutUser()` → `logoutPreviousUser()` → `logoutUser(withOnSuccess: nil, onFailure: nil)`. Smaller diff, original delegation preserved. Build green, AuthTests 42/0. Thanks for the review! 🙏

swift-sdk/SDK/IterableAPI.swift
joaodordio: Should this be updated with the same level of error as we have on logoutUser?

joaodordio (approved):

swift-sdk/SDK/IterableAPI.swift
sumeruchat: Good catch, addressed in 7708de1c. Both `disableDeviceForCurrentUser` and `disableDeviceForAllUsers` completion overloads now fire `onFailure("Iterable SDK is not initialized", nil)` on the not-initialized guard, matching `logoutUser`. The parameterless overloads pass nil handlers, so they are unchanged. Added a test, AuthTests 44/0.

PR #1071, opened 2026-05-29
SDK-469 Fix CodeQL string length conflation in deep link check

### LINKED ISSUE #794: CodeQL - String length conflation
Hi Team,

We've been running CodeQL scans regularly on our code base and have encountered an issue that I believe needs attention from the maintainers.

**Problem**

During our latest scan, CodeQL flagged an issue in file `Vault/Pods/Iterable-iOS-SDK/swift-sdk/Internal/DeepLinkManager.swift` at line number `96` with the following reason & description :

> **CodeQL**
>
> Vault/Pods/Iterable-iOS-SDK/swift-sdk/Internal/DeepLinkManager.swift:96
> **String length conflation**
> This String length is used in an NSString, but it may not be equivalent.
> 
> 
> 
> Using a length value from an NSString in a String, or a count from a String in an NSString, may cause unexpected behavior including (in some cases) buffer overwrites. This is because certain unicode sequences are represented as one character in a String but as a sequence of multiple characters in an NSString. For example, a 'thumbs up' emoji with a skin tone modifier (👍🏿) is represented as U+1F44D (👍) then the modifier U+1F3FF.
> 
> This issue can also arise from using the values of String.utf8.count, String.utf16.count or String.unicodeScalars.count in an unsuitable place.
> 
> Recommendation
> 
> Use String.count when working with a String. Use NSString.length when working with an NSString. Do not mix values for lengths and offsets between the two types as they are not compatible measures.
> 
> If you need to convert between Range and NSRange, do so directly using the appropriate initializer. Do not attempt to use incompatible length and offset values to accomplish conversion.
> 
> Example
> 
> In the following example, a String is converted to NSString, but a range is created from the String to do some processing on it.
> 
> 
> ```
> func myFunction(s: String) {
> 	let ns = NSString(string: s)
> 	let nsrange = NSMakeRange(0, s.count) // BAD: String length used in NSMakeRange
> 
> 	// ... use nsrange to process ns
> }
> ```
> 
> This is dangerous because, if the input contains certain characters, the range computed on the String will be wrong for the NSString. This will lead to incorrect behaviour in the string processing that follows. To fix the problem, we can use NSString.length to create the NSRange instead, as follows:
> 
> 
> ```
> func myFunction(s: String) {
> 	let ns = NSString(string: s)
> 	let nsrange = NSMakeRange(0, ns.length) // Fixed: NSString length used in NSMakeRange
> 	// ... use nsrange to process ns
> }
> 
> ```
> References
> 
> [Swift String vs. NSString](https://talk.objc.io/episodes/S01E80-swift-string-vs-nsstring)
> Common Weakness Enumeration: [CWE-135](https://cwe.mitre.org/data/definitions/135.html).
> 

This issue needs to be addressed to prevent it from being flagged in future scans.

**Request:**

Could the maintainers please review this issue. Our goal is to ensure that this particular CodeQL scan issue is resolved in upcoming scans to provide more security on the codebase.

**Additional Information**

We are using the cocoapod dependency manager for using Iterable with version-> `'Iterable-iOS-SDK', '6.5.4'`

Thank you for your attention to this matter.

sumeruchat:
## What
Fixes a CodeQL-flagged String-length conflation (CWE-135) in `DeepLinkManager.isIterableDeepLink`. The `NSRegularExpression` search range was built from `urlString.count` (a Swift `String` grapheme-cluster count) while `NSRegularExpression` operates on `NSString`/UTF-16 coordinates. For URLs with non-BMP characters (e.g. emoji), those measures diverge and the range under-covers the string, so a valid deep link can be missed. Ticket: [SDK-469](https://iterable.atlassian.net/browse/SDK-469).

## Changes
- `DeepLinkManager.swift`: build the `NSRange` from the bridged `NSString.length` instead of `String.count`, so the range matches the UTF-16 coordinates `NSRegularExpression` uses. Matches the ticket's prescribed fix and CodeQL's own recommended example.
- `DeepLinkTests.swift`: two regression tests — a URL with a long non-BMP prefix before the `/a/...` segment (this case returns `false` under the old `String.count` range and `true` with the fix), and a non-BMP URL without a deep-link path (stays `false`).

## Impact
- **Breaking changes**: None. Behavior only changes for previously-mishandled non-BMP URLs, which now match correctly.
- **Dependencies**: None.
- **Performance**: Negligible — one `String`→`NSString` bridge per call.

## Testing
**How to test:**
1. Run `./agent_test.sh DeepLinkTests` — 11/11 pass, including the two new SDK-469 cases.
2. Confirm CodeQL no longer flags the string-length conflation at `DeepLinkManager.swift`.
3. Edge case: a deep link like `https://links.iterable.com/👍🏿…/a/<id>` is now correctly recognized as an Iterable deep link.

[SDK-469]: https://iterable.atlassian.net/browse/SDK-469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swift-sdk/Internal/DeepLinkManager.swift
+++ b/swift-sdk/Internal/DeepLinkManager.swift
@@ -93,8 +93,13 @@ class DeepLinkManager: NSObject {
         guard let regex = try? NSRegularExpression(pattern: Const.deepLinkRegex, options: []) else {
             return false
         }
-        
-        return regex.firstMatch(in: urlString, options: [], range: NSMakeRange(0, urlString.count)) != nil
+
+        // NSRegularExpression operates on NSString (UTF-16) semantics, so the search
+        // range must be built from the NSString length, not String.count. Mixing the
+        // two conflates length measures and produces an incorrect range for strings
+        // containing non-BMP characters (e.g. emoji). See CWE-135 (SDK-469).
+        let nsString = urlString as NSString
+        return regex.firstMatch(in: urlString, options: [], range: NSRange(location: 0, length: nsString.length)) != nil
     }
     
     private lazy var redirectUrlSession: NetworkSessionProtocol = {
--- a/swift-sdk/Internal/DeepLinkManager.swift
+++ b/swift-sdk/Internal/DeepLinkManager.swift
@@ -94,12 +94,11 @@ class DeepLinkManager: NSObject {
             return false
         }
 
-        // NSRegularExpression operates on NSString (UTF-16) semantics, so the search
-        // range must be built from the NSString length, not String.count. Mixing the
-        // two conflates length measures and produces an incorrect range for strings
-        // containing non-BMP characters (e.g. emoji). See CWE-135 (SDK-469).
-        let nsString = urlString as NSString
-        return regex.firstMatch(in: urlString, options: [], range: NSRange(location: 0, length: nsString.length)) != nil
+        // Build the range with NSRange(_:in:) so it is in the UTF-16 coordinates
+        // NSRegularExpression expects. urlString.count conflates Swift String length
+        // with NSString length for non-BMP characters (CWE-135, SDK-469).
+        let range = NSRange(urlString.startIndex..<urlString.endIndex, in: urlString)
+        return regex.firstMatch(in: urlString, options: [], range: range) != nil
     }
     
     private lazy var redirectUrlSession: NetworkSessionProtocol = {
--- a/tests/unit-tests/DeepLinkTests.swift
+++ b/tests/unit-tests/DeepLinkTests.swift
@@ -149,7 +149,26 @@ class DeepLinkTests: XCTestCase {
     func testIsIterableDeepLinkReturnsFalseForEmptyString() {
         XCTAssertFalse(IterableAPI.isIterableDeepLink(""))
     }
-    
+
+    // SDK-469: when the URL contains non-BMP characters (e.g. emoji) BEFORE the
+    // deep-link path segment, String.count under-counts vs NSString.length (each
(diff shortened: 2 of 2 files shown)


swift-sdk/Internal/DeepLinkManager.swift
```diff
@@ -93,8 +93,13 @@ class DeepLinkManager: NSObject {
...
         }
-        
-        return regex.firstMatch(in: urlString, options: [], range: NSMakeRange(0, urlString.count)) != nil
+
+        // NSRegularExpression operates on NSString (UTF-16) semantics, so the search
+        // range must be built from the NSString length, not String.count. Mixing the
+        // two conflates length measures and produces an incorrect range for strings
+        // containing non-BMP characters (e.g. emoji). See CWE-135 (SDK-469).
+        let nsString = urlString as NSString
+        return regex.firstMatch(in: urlString, options: [], range: NSRange(location: 0, length: nsString.length)) != nil
```
joaodordio: We already use NSRange(string.startIndex..<string.endIndex, in: string) in UnknownUserManager+Functions.swift for the same problem. Worth aligning on one bridge style so we are not picking the dice each time. Not blocking the fix, just flagging the difference.

joaodordio (approved):

swift-sdk/Internal/DeepLinkManager.swift
sumeruchat: Done in 8edfa03b. Switched to `NSRange(urlString.startIndex..<urlString.endIndex, in: urlString)` to match `UnknownUserManager+Functions.swift`. DeepLinkTests 11/0.

PR #1073, opened 2026-06-05
SDK-470 Add CustomDebugStringConvertible to AuthFailureReason

sumeruchat:
## What
`AuthFailureReason` is an Int-backed `@objc` enum, so logging an instance prints a raw number. This makes it conform to `CustomDebugStringConvertible` so logs show the case name.

## Changes
- Add a `CustomDebugStringConvertible` extension mapping each case to its name.
- Switch is exhaustive with no `default`, so a future case fails to compile until labeled.
- Add a unit test covering all 11 cases.

## Impact
- **Breaking changes**: None. Additive only, public API and raw values unchanged.
- **Dependencies**: None.
- **Performance**: None.

## Testing
**How to test:**
1. Run `./agent_test.sh AuthTests`.
2. `testAuthFailureReasonDebugDescriptions` passes for all 11 cases.
3. `String(reflecting: AuthFailureReason.authTokenExpired)` returns `"authTokenExpired"`.

Jira: https://iterable.atlassian.net/browse/SDK-470

### DIFF
--- a/swift-sdk/Core/Utilities/AuthFailureReason.swift
+++ b/swift-sdk/Core/Utilities/AuthFailureReason.swift
@@ -20,3 +20,32 @@ import Foundation
     case authTokenGenerationError
     case authTokenMissing
 }
+
+extension AuthFailureReason: CustomDebugStringConvertible {
+    public var debugDescription: String {
+        switch self {
+        case .authTokenExpired:
+            return "authTokenExpired"
+        case .authTokenGenericError:
+            return "authTokenGenericError"
+        case .authTokenExpirationInvalid:
+            return "authTokenExpirationInvalid"
+        case .authTokenSignatureInvalid:
+            return "authTokenSignatureInvalid"
+        case .authTokenFormatInvalid:
+            return "authTokenFormatInvalid"
+        case .authTokenInvalidated:
+            return "authTokenInvalidated"
+        case .authTokenPayloadInvalid:
+            return "authTokenPayloadInvalid"
+        case .authTokenUserKeyInvalid:
+            return "authTokenUserKeyInvalid"
+        case .authTokenNull:
+            return "authTokenNull"
+        case .authTokenGenerationError:
+            return "authTokenGenerationError"
+        case .authTokenMissing:
+            return "authTokenMissing"
+        }
+    }
+}
--- a/tests/unit-tests/AuthTests.swift
+++ b/tests/unit-tests/AuthTests.swift
@@ -17,6 +17,26 @@ class AuthTests: XCTestCase {
     override func setUp() {
         super.setUp()
     }
+
+    func testAuthFailureReasonDebugDescriptions() {
+        let expectedDescriptions: [(AuthFailureReason, String)] = [
+            (.authTokenExpired, "authTokenExpired"),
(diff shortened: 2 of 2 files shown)


franco-zalamena-iterable (approved):

swift-sdk/Core/Utilities/AuthFailureReason.swift
```diff
@@ -20,3 +20,32 @@ import Foundation
     case authTokenGenerationError
     case authTokenMissing
 }
+
+extension AuthFailureReason: CustomDebugStringConvertible {
```
joaodordio, line 24: We already solve this exact problem for `IterableInAppContentType` and `IterableInAppTriggerType` in `InAppPersistence.swift` using `CustomStringConvertible` and `description`, with a comment explaining why it's needed for `@objc` enums. 

Should we stay consistent and use `CustomStringConvertible` + `description` here too? 

Both protocols technically work for `String(describing:)` thanks to the fallback chain, but `description` is the right semantic for output customers will see in their AuthDelegate logs, and it matches the existing convention in this SDK.

tests/unit-tests/AuthTests.swift
```diff
@@ -17,6 +17,26 @@ class AuthTests: XCTestCase {
...
+            (.authTokenInvalidated, "authTokenInvalidated"),
+            (.authTokenPayloadInvalid, "authTokenPayloadInvalid"),
+            (.authTokenUserKeyInvalid, "authTokenUserKeyInvalid"),
+            (.authTokenNull, "authTokenNull"),
+            (.authTokenGenerationError, "authTokenGenerationError"),
+            (.authTokenMissing, "authTokenMissing")
+        ]
+
+        expectedDescriptions.forEach { reason, description in
+            XCTAssertEqual(reason.debugDescription, description)
```
joaodordio: This is asserting `debugDescription` against the same strings we just hardcoded in the switch, so it can't fail in a meaningful way. 
The bug we're fixing is that `print(reason)` or `"\(reason)"` shows `AuthFailureReason(rawValue: 0)` instead of the case name. 

Can we switch this to `XCTAssertEqual("\(reason)", description)` (or `String(describing: reason)`) so the test actually covers the customer-facing logging path? 

That way if someone changes the protocol in the future and breaks `String(describing:)`, this test catches it.

swift-sdk/Core/Utilities/AuthFailureReason.swift
sumeruchat: Fair point, and the `InAppPersistence` precedent is useful context. Keeping `CustomDebugStringConvertible` here since SDK-470 asks for it specifically. It still covers the customer path: `print(reason)` and `"\(reason)"` go through `String(describing:)`, which falls back to `debugDescription`, so AuthDelegate logs show the case name. Updated the test (your other comment) to assert `String(describing: reason)` so it guards that path.

tests/unit-tests/AuthTests.swift
sumeruchat: Good call. Switched the assertion to `String(describing: reason)` so it covers the `print(reason)` / `"\(reason)"` path instead of `debugDescription` directly.

PR #1074, opened 2026-06-08
SDK-291 Fix in-app message position in landscape

sumeruchat:
## What
In-app HTML messages could be mispositioned after rotating to landscape on notched devices. The webview width and center were computed from the full parent width, ignoring the side safe-area inset where the Dynamic Island or notch sits.

## Changes
- Constrain webview width and `center.x` using `safeAreaInsets.left` and `.right`, so a message stays inside the safe area in any orientation.
- Center a `.center` message within the top and bottom safe-area bounds.
- Make the location switch exhaustive (explicit `.center` and `.full`) so a future case fails to compile until handled.
- Pure inset-based math, no orientation or global-state reads. Added unit tests for portrait and both landscape orientations.

## Impact
- **Breaking changes**: None. Internal layout math only, no API change.
- **Dependencies**: None.
- **Performance**: None.

## Testing
**How to test:**
1. Run `./agent_test.sh IterableHtmlMessageViewControllerTests` (34 tests, including new portrait and landscape cases).

**Visual check still needed (not covered by unit tests):** on a notched device or simulator, show a top, bottom, and center HTML in-app message, rotate to both landscape orientations, and confirm the message stays inside the safe area and clear of the Dynamic Island and home indicator. The unit tests verify the inset math but cannot confirm live rotation behavior or how customer HTML renders in the adjusted frame.

Jira: https://iterable.atlassian.net/browse/SDK-291

### DIFF
--- a/swift-sdk/Internal/in-app/InAppCalculations.swift
+++ b/swift-sdk/Internal/in-app/InAppCalculations.swift
@@ -111,14 +111,15 @@ struct InAppCalculations {
         position.height = inAppHeight
         
         // now set the width
+        let availableWidth = parentPosition.width - safeAreaInsets.left - safeAreaInsets.right
         let notificationWidth = 100 - (paddingLeft + paddingRight)
-        position.width = parentPosition.width * notificationWidth / 100
+        position.width = availableWidth * notificationWidth / 100
         
         // Position webview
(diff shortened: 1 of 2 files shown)


joaodordio (approved):
LGTM

Issue #275, opened 2020-03-03
IterableHtmlMessageViewController with no content and blocking interaction with the app


osawhoop:
Hello Iterable Team,
I noticed a bug in the app that when a button has no "action" or "url" is empty the SDK add a IterableHtmlMessageViewController on top of the hierarchy blocking any interaction with the app, the user needs to force quit the app in order to continue using it, I wonder if could you add some validation in the SDK to prevent this behavior.

this is an example of the template I'm using:
`<a class="btn" href="">LET ME SEE !</a> <a class="btn2" href="action://dismiss">I'M GOOD</a>`
the button "I'M GOOD" close the in-app message but the button "LET ME SEE!" trigger to show the IterableHtmlMessageViewController with no content blocking the interaction with the app. I'm using the SDK 6.2.0

Thanks! and keep doing a great job with this SDK! we love it!

<img width="379" alt="Screen Shot 2020-03-03 at 9 09 08 AM" src="https://user-images.githubusercontent.com/43786155/75784152-c77e2300-5d2f-11ea-8871-d80e5677e5af.png">

tapashmajumder:
Sorry our SDK does not support buttons with blank hrefs.

Our UI does not even allow you to save an in-app without at least one valid href link. So, in this case, if you remove the "I'M GOOD" button, you won't be able to save this in-app.

osawhoop:
oh I did not know about that, thank you!

PR #1079, opened 2026-06-25
SDK-490 Log in-app HTML render failures for Android parity

sumeruchat:
## What
The iOS SDK logged nothing when an in-app HTML message failed to render. This adds that logging and dismisses the in-app if its web content process dies, matching the Android SDK (SDK-490).

## Changes
`IterableHtmlMessageViewController` implemented only `WKNavigationDelegate`'s `didFinish` and `decidePolicyFor`. This adds the failure callbacks:
- `didFail` and `didFailProvisionalNavigation` log the failure and message id via `ITBError`.
- `webViewWebContentProcessDidTerminate` logs, then dismisses through the existing `createDismisser` so a dead webview does not leave a blank view blocking the app.
- A test covers the content-process-termination dismiss.

## Android parity scope
Android hardened in-app HTML across SDK-368 (webview-creation crash), SDK-419 (oversized payload, `TransactionTooLargeException`), and SDK-416 (renderer crash, `onRenderProcessGone`). Only the last translates to iOS: `onRenderProcessGone` maps to `webViewWebContentProcessDidTerminate`, which the SDK did not handle, and it is also the closest analog to a too-large message (memory pressure killing the content process). `TransactionTooLargeException` (an Android Binder limit) and the webview-creation `Resources.NotFoundException` have no iOS equivalent, so they are out of scope. No iOS GitHub issue reports this class; #275 (blank in-app from an empty `href`) is the closest, which this dismiss also guards.

## Impact
- **Breaking changes**: None. No public or Obj-C API change; the logging is additive.
- **Behavior**: a terminated web content process now dismisses the in-app instead of leaving it blank and unresponsive.

## Testing
`./agent_test.sh IterableHtmlMessageViewControllerTests` passes (35 tests, including the new `testWebContentProcessTerminationDismissesNonModalInApp`). A render failure logs `Unable to render in-app HTML for messageId: <id>, ...`.

A full `./agent_test.sh` flakes on `offline-events-tests/TaskRunnerTests/testMixedQueueOnlyUnauthenticatedExecuteDuringAuthPause`, a known timing-flaky test unrelated to this change.

Jira: https://iterable.atlassian.net/browse/SDK-490


tests/unit-tests/IterableHtmlMessageViewControllerTests.swift
```diff
@@ -503,6 +503,19 @@ class IterableHtmlMessageViewControllerTests: XCTestCase {
         let webViewDidFinishLoadingAfter = mirrorAfter.children.first(where: { $0.label == "webViewDidFinishLoading" })?.value as? Bool
         XCTAssertEqual(webViewDidFinishLoadingAfter, true, "webViewDidFinishLoading should be true after didFinish is called")
     }
+
+    func testWebContentProcessTerminationDismissesNonModalInApp() {
```
joaodordio, line 510: The new test covers isModal: false, but real in-apps come through InAppDisplayer with isModal: true, which routes to dismiss(animated:), not popViewController. Can we add a test that asserts the modal dismiss path is invoked too, ideally for both isInboxMessage: true and false so the animated: argument is also pinned down? Without it the primary in-app behavior this PR claims to fix is untested.

swift-sdk/Internal/IterableHtmlMessageViewController.swift
```diff
@@ -392,6 +396,21 @@ extension IterableHtmlMessageViewController: WKNavigationDelegate {
...
+
+    func webView(_: WKWebView, didFail _: WKNavigation!, withError error: Error) {
+        ITBError("Unable to render in-app HTML for messageId: \(messageIdForLogging), navigation failed with error: \(error)")
+    }
+
+    func webView(_: WKWebView, didFailProvisionalNavigation _: WKNavigation!, withError error: Error) {
+        ITBError("Unable to render in-app HTML for messageId: \(messageIdForLogging), provisional navigation failed with error: \(error)")
+    }
+
+    func webViewWebContentProcessDidTerminate(_: WKWebView) {
```
joaodordio, line 413: When webViewWebContentProcessDidTerminate dismisses, viewWillDisappear still runs and fires trackInAppClose with source: .link for modal in-apps (the common case) or .back for inbox nav. 

That attributes every render-process kill to a fake user action and skews customer analytics. 

Can we set a flag on this dismiss path and skip the trackInAppClose in viewWillDisappear when the dismiss is render-failure driven? Or maybe we should add a new InAppCloseSource case like .error, to accuratley track what's actually happenning.

joaodordio (changes requested):
We need to decide what to do when webViewWebContentProcessDidTerminate dismisses, because viewWillDisappear still runs and fires trackInAppClose which is inaccurate since the InApp failed to render and was not closed by user action.

joaodordio:
Another note: 

CHANGELOG.md is not updated. This PR introduces new behavior and logging worth a changelog line. Add an entry under [Unreleased] / Fixed describing the new dismiss-on-content-process-termination behavior and the failure logging.

Minor: didFail and didFailProvisionalNavigation are uncovered by tests. A small assertion via a mock IterableLogDelegate (the SDK already has one) would lock the log message format.

swift-sdk/Internal/IterableHtmlMessageViewController.swift
sumeruchat: Good catch. `webViewWebContentProcessDidTerminate` now sets a `dismissedDueToRenderFailure` flag, and `viewWillDisappear` returns before `trackInAppClose` when it is set, so a content-process kill no longer logs a fake `.link`/`.back` close. I went with the flag rather than a new `InAppCloseSource.error` case to avoid a public `@objc` enum change plus backend and reporting coordination; the failure is still captured by the `ITBError` log. Locked with `testRenderFailureDismissDoesNotTrackInAppClose`.

tests/unit-tests/IterableHtmlMessageViewControllerTests.swift
sumeruchat: Done. Added `testCreateDismisserDismissesModalInApp`, which drives the modal branch of `createDismisser` directly and asserts `dismiss(animated:)` is called with `false` for non-inbox and `true` for inbox. Used a small dismiss-recording `UIViewController` rather than swizzling. The VC-to-dismisser wiring stays covered by the existing non-modal termination test.

sumeruchat:
Both addressed. Added a CHANGELOG entry under Unreleased / Fixed for the render-failure logging and the dismiss-on-content-process-termination behavior. And added `testDidFailLogsRenderFailure` and `testDidFailProvisionalNavigationLogsRenderFailure`, which assert the error level and message format via a mock `IterableLogDelegate`.

joaodordio (approved):

PR #1080, opened 2026-07-06
SDK-498 Add unknown user criteria fetch callbacks

sumeruchat:
## What
Adds two callbacks to `IterableUnknownUserHandler` so apps can react when unknown-user activation criteria are fetched, matching the Android SDK (SDK-498). Apps often need to update user data right after init, but the criteria are fetched asynchronously and are required to create the unknown user, so these callbacks give a hook once the criteria arrive.

## Changes
- Add `@objc optional func onCriteriaReceived(criteria: [AnyHashable: Any])` and `@objc optional func onCriteriaFetchFailed(reason: String)` to `IterableUnknownUserHandler`. Optional to mirror Android's `default` methods, so existing conformers are unaffected.
- `UnknownUserManager.getUnknownCriteria()` persists the fetched criteria first, then fires `onCriteriaReceived` (only when the payload parses), matching Android's guarantee that track calls made from the callback evaluate against the just-arrived criteria. Fires on every fetch.
- `onCriteriaFetchFailed(reason:)` fires on a failed fetch; `getCriteriaData` now surfaces the failure reason.

## Impact
- **Breaking changes**: None. The new protocol methods are `@objc optional`, additive.
- **Parity**: matches the merged Android implementation. Android PR 1058 was closed; the version that shipped added these methods to the existing handler.

## Testing
**How to test:**
1. Run `./agent_test.sh IterableApiCriteriaFetchTests` (includes the two new tests).
2. A successful criteria fetch fires `onCriteriaReceived` with the parsed criteria, with storage already populated when the callback runs. A failed fetch fires `onCriteriaFetchFailed(reason:)` and does not write criteria to storage.

Jira: https://iterable.atlassian.net/browse/SDK-498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
+### Added
+- Added optional `onCriteriaReceived(criteria:)` and `onCriteriaFetchFailed(reason:)` callbacks to `IterableUnknownUserHandler` for unknown user activation criteria fetch results.
+
 ### Fixed
 - In-app HTML render failures now log the message ID and error, and messages are dismissed if the WKWebView content process terminates.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
-- Added optional `onCriteriaReceived(criteria:)` and `onCriteriaFetchFailed(reason:)` callbacks to `IterableUnknownUserHandler` for unknown user activation criteria fetch results.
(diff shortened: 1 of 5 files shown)


joaodordio (approved):
LGTM!

A few small things I'd love to see, none of them blocking:

- Could you flesh out the CHANGELOG entry a bit? Android's version mentions that callbacks fire on the main thread and may fire on every fetch (init, foreground, when visitor usage tracking gets enabled), so implementations should handle repeated invocations. The protocol doc-comment already says "Fires on every criteria fetch" but the main-thread guarantee isn't in any Swift-facing doc.
- One more test to pin the "fires on every fetch" contract would be great, similar to Android's IterableApiCriteriaFetchTests.java that asserts receivedCount == 2 across two fetches.
- Optional, but the JSONSerialization.data(withJSONObject:) catch in getCriteriaData still just prints and now silently drops both completion and onFailure. Since a handler is subscribed now, it'd be nice if that path also called onFailure (or at least ITBError) so parse-side failures don't disappear.

sumeruchat:
Addressed all three. One note: the main-thread mention turned into a small code change. Callbacks were firing on the URLSession completion queue, so both handler calls are now dispatched to main to match Android. The new test pins the two-fetch contract and main-thread delivery.

PR #1085, opened 2026-07-20
SDK-562 Fix offline queue replaying expired JWT

sumeruchat:
## What
Requests persisted to the offline queue froze the JWT at enqueue time and replayed it on every retry. A task queued while the token was expired kept failing with a 401 even after a successful refresh, survived app restarts, and paused the JWT tasks queued behind it. Reported by Wolt ([SDK-562](https://iterable.atlassian.net/browse/SDK-562)).

## Changes
- `IterableAPICallTaskProcessor` overwrites the decoded task's auth token with the current `AuthManager` token before building the request, matching the online path which reads the token at send time
- `IterableTaskRunner` and the dependency container thread the existing `AuthManager` instance into the processor (optional parameter, nil default, so other constructions are unchanged)
- A nil live token keeps the persisted one, and the persisted task format is unchanged, so tasks already stuck on devices heal on their next run after a refresh
- Tests reproduce the poisoned-task loop end to end: 401 with the expired token, refresh, retry goes out with the fresh token, queue drains. Re-stamp semantics across a process restart and a multi-task drain variant are also covered

## Impact
- **Breaking changes**: None
- **Dependencies**: None
- **Performance**: One `getAuthToken()` read per offline task execution

## Testing
**How to test:**
1. `./agent_build.sh`
2. `xcodebuild test -scheme swift-sdk -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -only-testing:offline-events-tests` (96 tests, 0 failures; full unit-tests also green, 644 tests)
3. Key tests: `TaskRunnerTests.testJWTAuthFailureRetriesWithRefreshedToken` asserts the wire sees `Bearer expired-token` then `Bearer fresh-token`; `TaskProcessorTests.testUsesCurrentAuthTokenWhenProcessingPersistedTask` pins re-stamp across a restart
4. Red-test proof: with the re-stamp reverted, the loop test fails with headers `[expired, expired]`, confirming it pins the reported bug

[SDK-562]: https://iterable.atlassian.net/browse/SDK-562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
+### Fixed
+- Fixed offline-queued requests replaying an expired JWT forever. Tasks persisted while the token was expired kept the stale token in their payload, so they failed with a 401 on every retry even after a successful refresh, and could block the rest of the offline queue. The task processor now stamps the current auth token at execution time, matching online behavior, which also heals tasks already stuck in the queue.
 
 ## [6.7.4]
 ### Added
--- a/swift-sdk/Internal/IterableAPICallRequest.swift
+++ b/swift-sdk/Internal/IterableAPICallRequest.swift
@@ -9,7 +9,7 @@ import Foundation
 struct IterableAPICallRequest {
     let apiKey: String
     let endpoint: String
-    let authToken: String?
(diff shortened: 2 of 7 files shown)


joaodordio (approved):
LGTM

PR #1086, opened 2026-07-20
SDK-569 Parse itbl delivered as a JSON string (iOS-via-FCM)

sumeruchat:
## What
iOS devices registered via FCM receive pushes where `userInfo["itbl"]` is a JSON string, not a dictionary, because FCM's `message.data` is a `Map<String, String>` and gets merged into the APNs payload. Every SDK parse site did `as? [AnyHashable: Any]`, which returns nil for a string, so push-open tracking, ghost/silent classification, default actions, deep links, action buttons, and notification-extension rich media all silently no-op. This adds tolerant `itbl` extraction so those flows work whether `itbl` arrives as a dictionary or a string.

Found during MOB-10018 (FCM analytics_label) staging validation, and confirmed on the legacy FCM path too, where there is no `apns.payload` object to fall back on, so this SDK change is the only possible fix on that path.

## Changes
- Tolerant `itbl` extraction: a dictionary is used as-is; a `String` is JSON-decoded via `JSONSerialization`; malformed input returns nil rather than crashing.
- Applied at all four parse sites: `NotificationHelper` and `InternalIterableAppIntegration` (IterableSDK target), `NotificationContentParser` and `ITBNotificationServiceExtension` (IterableAppExtensions target). The two targets keep separate helpers because they are independent SPM targets with no shared dependency.
- Tests at both the parse layer and the flow layer: open tracking, custom action, and attachment retrieval all driven from a string `itbl`, plus malformed-string handling.

## Impact
- **Breaking changes**: None. Dictionary payloads (direct APNs) parse exactly as before; the string path is purely additive.
- **Dependencies**: None.
- **Performance**: one extra JSON decode, and only when `itbl` arrives as a string (the FCM path). The dictionary path is unchanged.

## Testing
**How to test:**
1. Register an iOS device via FCM (platform GCM / firebaseJson) so pushes arrive with a stringified `itbl`.
2. Send a push and open it: the open tracks (campaign / template / message id), default actions and deep links fire, action buttons render, and rich-media attachments load.
3. Confirm direct-APNs (dictionary `itbl`) behavior is unchanged.

**What I tested:**
- [x] `unit-tests/NotificationMetadataTests`: valid and malformed string `itbl` parsing
- [x] `unit-tests/NotificationResponseTests`: open tracking and custom action fired from a string `itbl`
- [x] `notification-extension-tests/NotificationExtensionTests`: action buttons and image attachment from a string `itbl`, malformed string ignored

[SDK-569]: https://iterable.atlassian.net/browse/SDK-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/notification-extension/ITBNotificationServiceExtension.swift
+++ b/notification-extension/ITBNotificationServiceExtension.swift
@@ -52,7 +52,7 @@ public protocol ITBNotificationServiceExtensionDelegate {
     // MARK: - Private
     
     private func retrieveAttachment(from content: UNNotificationContent) {
-        guard let metadata = content.userInfo[JsonKey.Payload.metadata] as? [AnyHashable: Any],
+        guard let metadata = NotificationContentParser.getIterableMetadata(from: content),
               let attachmentUrlString = metadata[JsonKey.Payload.attachmentUrl] as? String,
               let url = URL(string: attachmentUrlString) else {
             attachmentRetrievalFinished = true
--- a/notification-extension/NotificationContentParser.swift
+++ b/notification-extension/NotificationContentParser.swift
@@ -7,7 +7,7 @@ import UserNotifications
 
 struct NotificationContentParser {
     static func getIterableMetadata(from content: UNNotificationContent) -> [AnyHashable: Any]? {
-        content.userInfo[JsonKey.Payload.metadata] as? [AnyHashable: Any]
+        getIterableMetadata(from: content.userInfo)
     }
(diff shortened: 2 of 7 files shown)


joaodordio (approved):
LGTM

PR #1087, opened 2026-07-21
SDK-496 Fix inAppConsume behavior and add JSON-only message signal

sumeruchat:
## What

This PR addresses [SDK-496](https://iterable.atlassian.net/browse/SDK-496) by fixing local in-app consumption and inbox notifications, adding durable JSON-only availability, and hardening identity switching.

The design and delivery contract are documented here: https://iterable.slab.com/posts/changes-to-in-app-inbox-notifications-and-json-only-message-behavior-ios-gvzb5sco

## Changes

- Public `inAppConsume` now removes the message locally, sends the consume request, and posts `iterableInboxChanged` after local state changes only for inbox messages. Add and remove paths ignore non-inbox messages, and silent-push removal persists the post-removal list.
- JSON-only messages are saved to UserDefaults before the main-thread delegate method `onJsonOnlyMessageAvailable(message:)` and notification `iterableJsonOnlyInAppMessageAvailable` fire. `getUnhandledJsonOnlyMessages()` recovers pending messages, which replay on foreground until `markJsonOnlyMessageHandled(messageId:)` acknowledges them.
- The per-user queue clears on logout or user switch, honors `expiresAt`, and applies 30 day retention, 100 messages per user, and immediate triggers only. Those limits are SDK implementation defaults on a local queue, not a public contract.
- An acknowledged ID with an unchanged payload stays suppressed while its fingerprint record is retained, a changed payload is delivered as a new record, and an unacknowledged expired or evicted ID is not readmitted while its discard record is retained. The latest 100 acknowledgements and 100 discard records per user are retained, oldest forgotten first. This state applies only to JSON-only messages, so HTML reuse of an ID is unaffected.
- `IdentityCoordinator` scopes fetch, delivery, queued processing, and auth reads to the identity that started them. Stale work is discarded, and within a running process public getters do not return messages from another user.

## Impact

There are no API surface breaking changes and no new dependencies. The six intentional behavior changes are:

- `inAppConsume` callers now get the documented local removal.
- `iterableInboxChanged` no longer fires for non-inbox additions, removals, or read-marking, so showing a popup no longer announces an inbox change.
- `onNew` for a JSON-only message now fires asynchronously on the main thread, at most once, instead of synchronously during fetch processing.
- In a mixed fetch, JSON-only messages are delivered in server arrival order before HTML display selection, while HTML keeps priority ordering.
- JSON-only availability is delivered while auto-display is paused or during popup cooldown because those gates apply only to HTML.
- A callback already in flight for the previous user may complete after a concurrent identity switch; the SDK revalidates when it returns and performs no further delivery step for it. Callbacks hold no SDK lock, so they may call any SDK API from any thread.

Eligible queue insertion uses one UserDefaults write per fetch batch, first-delivery state persists once per record, and the JSON store is consulted on every identified-user fetch; HTML rendering paths are otherwise unchanged.

Known pre-existing limitation: the normal in-app list on disk has no identity label, so a hard process kill between a user switch and queued cleanup can make the previous list load as the current user until the next sync; in-process reads are identity-gated, and the identity-labeled JSON-only queue does not have this gap. Scope is frozen at this revision; further non-regression findings are tracked as follow-up tickets.

## Testing

1. Run `./agent_build.sh`.
2. Run `./agent_test.sh`.

The local full run is green with unit 693, notification-extension 12, offline-events 96, and 0 failures; CI can flake on three unrelated SDK-99 asynchronous `disablePush` tests that predate this change.

Fix tests verify post-mutation notification ordering and no notification for non-inbox removal. A 37-test `InAppTests` suite covers persistence before signaling, ordering, fingerprint acknowledgement and readmission, cold-start recovery, replay, deduplication, identity races, display gates, cross-type ID reuse, consume failures, expiry, retention, capacity, tombstones, and batch persistence. Objective-C compile tests cover both the legacy conformer and the new API surface.

[SDK-496]: https://iterable.atlassian.net/browse/SDK-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


swift-sdk/Internal/in-app/InAppManager.swift
```diff
@@ -506,6 +681,149 @@ class InAppManager: NSObject, IterableInternalInAppManagerProtocol {
...
+            }
+            guard self.identityCoordinator.performIfCurrent(identityContext,
+                                                            identityProvider: self.identityProvider, {
+                self.notificationCenter.post(name: .iterableJsonOnlyInAppMessageAvailable,
+                                             object: delivery.message,
+                                             userInfo: nil)
+            }) else {
+                result.resolve(with: false)
+                return
+            }
```
joaodordio, line 776: These callbacks run while holding the identity lock, which is a deadlock risk.

`onNew`, `onJsonOnlyMessageAvailable`, and the notification post are passed as the block to `performIfCurrent`, which executes that block inside `withCriticalSection`'s `NSRecursiveLock` (`Auth.swift:33`, `61-65`, `75`), on the main thread (`InAppManager.swift:790-794`). So the customer callback runs with the identity lock held.

`IterableAPI.email` and `auth` (which most `track`/network calls build their request from) re-enter that same lock through `identityValues()` (`InternalIterableAPI.swift:21`, `81-82`, `975-977`). So a callback that hops to another queue and synchronously waits while that work reads an identity API deadlocks: the worker thread blocks on the lock main holds, and main is blocked on the sync. Same-thread reentry is fine thanks to the recursive lock, but a cross-thread synchronous wait is an easy pattern to hit in a callback whose whole job is to hand the customer a payload to process.

Can we snapshot what we need under the lock, release it, invoke the customer callback outside the critical section, then re-validate identity (generation + snapshot) after it returns? If we keep it as-is, the restriction needs to be a lot louder in both the delegate header and docs PR

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -941,8 +969,18 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
...
     
     private func retrieveIdentifierData() {
-        _email = localStorage.email
-        _userId = localStorage.userId
+        setIdentity(email: localStorage.email, userId: localStorage.userId)
+    }
+
+    private func identityValues() -> (email: String?, userId: String?) {
+        identityCoordinator.withCriticalSection { (_email, _userId) }
+    }
```
joaodordio, line 980: This is the re-entry point that closes the deadlock loop: `identityValues()` takes the same identity lock, so any identity read (`email`, `auth`, and `track` via `auth`) called from a callback-spawned thread while the JSON-only callback holds the lock on main will block here. See the callback-under-lock comment in `InAppManager.swift:754`.

joaodordio (changes requested):
The code quality and test depth are genuinely very good! 

I'm marking as CR so we can make a clear decision on the In AppManager with the callback-under-lock design!

swift-sdk/Internal/in-app/InAppManager.swift
sumeruchat: Adopted in 6200375b. The three surfaces now run with no SDK lock held: check identity before, invoke unlocked, revalidate on return; a stale result stops everything after that surface, and the mutation/consume guard is unchanged. The cross-thread wait restriction is gone from the public docs, replaced by the actual contract: a callback already selected for the previous user may complete after a concurrent switch, and no later SDK step runs for it.

swift-sdk/Internal/InternalIterableAPI.swift
sumeruchat: Resolved by the same change (6200375b): callbacks no longer hold the identity lock, so this re-entry can no longer close a deadlock loop.

swift-sdk/Internal/in-app/InAppManager.swift
sumeruchat: Good catch btw!

tests/unit-tests/InboxTests.swift
joaodordio, line 485: This test is failing on CI with count 2 on the first `iterableInboxChanged` callback. The new fetch pipeline resolves the Pending after `finishSync` posts to `callbackQueue`, and this observer then hops to `main.async` before reading `getInboxMessages()`. That lets payload 2 land before the first callback runs, so the first fulfillment sees both inbox messages. Either wait for the first inbox callback (and the first `onNew`) before sending payload 2, or assert on the callback queue without the extra `main.async`. 

Let's get this green before merge since it sits on the path this PR reworked.

joaodordio (commented):
Thanks for addressing the changes! Much better now.

Let's get the CI to pass green on all tests specifically on the one I've commented on.

Can we please also fix these: 
- Remove or resolve `// Product defaults pending confirmation.` for the 30-day / 100 caps.
- `set(read:)` still posts `iterableInboxChanged` for any message. Minor inconsistency with the new inbox-only rule.

sumeruchat:
`8003bedf` added JSON-only replay to `start()` reading `applicationStateProvider.applicationState` directly. Off-main that is an invalid `UIApplication.applicationState` read, and Xcode 16.x runs async XCTest bodies on a concurrency worker, so the three async `disableDevice` tests terminated the test host: no assertion, run restarts mid-class. Fixed in `14596663` by resolving the state through a main-thread helper, mirroring `getAppIsReady` without its display gate. The regression test asserts the read happens on main, so it fails deterministically rather than by scheduling luck. It never reproduced locally because CI pins Xcode 16.4 and I was on 26.6.

`InboxTests.testInboxAndInAppCallbacksTogether` was separate and pre-existing: no happens-before edge between the first inbox notification and the second payload, so the deferred observer read the later count. Hardened in the same commit, no assertion weakened. CI is green on `14596663`, both required checks.

Your two follow-ups are in `10113ae7`:
- `set(read:)` now posts `iterableInboxChanged` only for `saveToInbox` messages, so showing a popup no longer announces an inbox change. Tests cover both directions. Note this is a behavior change beyond what the PR claimed before, so the design doc now lists read-marking as gated; `reset()` and sync overwrite still post unconditionally.
- Dropped the provisional marker on the 30 day / 100 record limits. They are SDK limits on a local queue rather than a public contract, and the doc and PR body say so.

joaodordio (approved):

PR #1091, opened 2026-08-25
SDK-674 Add switchProject for runtime project switching

joaodordio:
# 📝 Summary
> Adds `IterableAPI.switchProject(project:callback:)`, moving a running install between Iterable projects in place with no restart and no state leaking across.

## 📖 Description

Multi-region apps have no way today to move a running install from one Iterable project to another without an app restart, which forces either a relaunch or separate builds per region. This adds a single entry point that tears the outgoing project down in place and stands up a replacement instance against the new key and config.

The sequence disables the push token on the outgoing project, resets its in-app and embedded state, purges its persisted offline queue while preserving queued `disableDevice` tasks, clears its identity from memory and storage along with the rest of its project-scoped storage (activation criteria, unsent unknown-user events and sessions, stored attribution), releases the outgoing instance so its `deinit` runs, and re-initializes. The method returns immediately and does the work off the calling thread.

**Three decisions worth reviewer attention.**

*Not everything is queued.* Calls made during the switch window are held and replayed, but only where replaying against the new project is the desired outcome. Anything carrying the outgoing project's campaign, template or message IDs runs against the project the SDK is on when it is made, because replaying it would attribute it to a project those IDs do not exist in. Push handling is deliberately outside the gate for that reason, and because holding the OS completion handler for the length of a switch risks the watchdog. `IterableAppIntegration`'s behaviour is unchanged for apps that never call `switchProject`. The gated set matches Android: `setEmail`, `setUserId`, `updateEmail`, `updateUser`, `updateCart`, `trackPurchase`, `track`, `registerToken`, `logoutUser` and the `email`/`userId` setters.

*The teardown waits for the disable handoff.* A request built with the outgoing API key cannot be sent once its instance is gone, and `OfflineRequestProcessor` holds its auth provider weakly, so releasing the instance before the disable reached the request layer dropped it silently without reporting a failure, leaving the device registered on the project the app had left. The teardown now awaits the handoff, meaning persisted to the queue or in flight online, with a bound matching Android's `DISABLE_DISPATCH_TIMEOUT_MS`. It deliberately does not await the network response, which in offline mode may never arrive.

*`attributionInfo` refuses late writes.* The setter drops writes once the project has been switched away from. Work the outgoing instance owns can resolve after the clear, a universal link redirect returning being the obvious case, and writing it back would attribute the new project's first event to a campaign it has never heard of.

The callback is delivered on the main thread. `true` means every cleanup step completed and a device disable was confirmed. `false` means the SDK **is** on the new project but a step was noisy; it never means the switch failed or rolled back, and it is the expected result for an app that does not use push. `switchProject` does not re-identify the user; call `setEmail`/`setUserId` from the callback.

Paired with [SDK-673](https://iterable.atlassian.net/browse/SDK-673) on Android, which implements the same contract. Divergence between the two is a bug, so the gated call set, the `false` conditions and the dispatch bound were all aligned deliberately.

## 🧪 How to test?

Full suite green on `iPhone 17 Pro Max`: **837 tests, 0 failures** (719 unit-tests, 15 notification-extension-tests, 103 offline-events-tests).

New coverage:

- `SwitchProjectTests` (17): identity and storage clearing, manager rebuild, instance deallocation, config swap, FIFO drain, rapid and repeated switches, no-op on unchanged key, callback thread, `false` conditions.
- `SwitchProjectGateTests` (8): a push tapped mid-switch is not handed to the new project, the silent-push completion handler is not held, the gate holds identity while releasing project-scoped calls, disable handoff signalling and its bound, attribution not written back after teardown, a failed re-init reporting `false`, empty and blank API key rejection.
- `SwitchProjectOfflineTests`: the teardown waits for the disable to reach the queue, and a queued disable survives the switch still targeting the outgoing project.
- `RequestHandlerTests/testDeleteAllTasksOnLogout` was stale (asserting `count == 0` against the preserve-`disableDevice` behaviour from SDK-297) and sat in `skippedTests`. Fixed and re-enabled.

Manually: initialize against project A, identify a user, register for push, then `switchProject` to project B. Confirm the device is disabled on A, that B has no trace of A's identity, criteria, unknown-user events or attribution, and that calls made during the window land on B in order.

## 🧾 Changelog

Added, see `CHANGELOG.md`. Documents the callback contract, the conditions producing `false`, which calls are queued versus run inline, and the disable handoff wait.

## 📹 Loom recording if applicable

N/A.

#### 🐞 Github Issues solved

None.

#### 📚 Docs PR if applicable

Required, tracked by [SDK-676](https://iterable.atlassian.net/browse/SDK-676) for both platforms. Not yet opened.

---

**Known, not addressed here**

- A purged offline task's `Pending` never resolves, so `setEmail`'s completion handler is stranded when a queued register is discarded by a logout purge. Android fixes this in SDK-675; iOS has the same gap and it predates this change.
- `unit-tests/testCriteriaUserIdTokenCheck` is load-sensitive rather than flaky in the usual sense, and unrelated to this change. It waits 5 seconds inside a 6 second timeout, so it has under a second of headroom. On an unloaded machine it passes at 5.17s; on a run where the unit target took 1242s instead of 296s it exceeded the bound and failed. Worth widening that bound or removing the fixed wait, separately from this PR.

[SDK-674]: https://iterable.atlassian.net/browse/SDK-674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-673]: https://iterable.atlassian.net/browse/SDK-673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

## The project pair (added after review feedback)

`switchProject` now takes an `IterableProject`, which pairs a project's API key with the `IterableConfig` to run it with:

```swift
guard let euProject = IterableProject(apiKey: euApiKey, config: config) else { return }
IterableAPI.switchProject(project: euProject) { cleanTeardown in ... }
```

`IterableConfig` carries `dataRegion` and `authDelegate`, so a loose key and config allowed one project's key to be combined with another project's region or auth delegate. That mismatch surfaces as an auth failure, or a request sent to the wrong region, that looks unrelated to the project it came from. Pairing them makes it unrepresentable.

The initializer is failable and returns `nil` for an empty or whitespace-only key, so an unusable project is not constructible. Android throws `IllegalArgumentException` for the same case: the guarantee is identical, the mechanism follows each platform's convention, since a blank key is a normal outcome of a failed region lookup and trapping on it would be wrong.

The `apiKey`/`config` form is now internal, used by the tests that inject a dependency container. Replaced rather than overloaded because this is unreleased; keeping both public would leave the loose form available forever.

**One caveat documented on the type:** `IterableConfig` holds its delegates `weak`, and this does not change that. A paired value object invites being held as a `static let`, so if the auth delegate is only retained by the config it deallocates and auth stops working silently. Android holds its handlers strongly and does not have this problem. This is the one way the abstraction makes something more dangerous rather than less, so it is called out in the doc comment and in the customer guide.

Also corrected in this branch: the JWT retry budget note previously claimed an exhausted budget could suppress the new project's first token request. It cannot. The retry state is private instance state on `AuthManager` and the switch builds a fresh dependency container, so the new project starts clean. Comment and CHANGELOG only.

---

## The result type (added after review feedback)

The callback reports an `IterableProjectSwitchResult`, either `.switchedCleanly` or `.switchedWithWarnings`, rather than a boolean.

The boolean was the problem the customer guide kept having to write around: `false` did not mean the switch failed, did not mean it was rolled back, was the expected outcome for any app without push, and called for exactly the same handling as `true`. A value needing that much explanation to avoid being misread is the wrong shape. Both case names lead with "switched" so the part that holds in either case, that the SDK is on the new project, is visible at the call site instead of being something the reader has to recall.

Declared `@objc` with an `Int` raw value, so Objective-C callers get `IterableProjectSwitchResultSwitchedCleanly` and friends. That rules out associated values, which is the main reason there is no warning detail attached to the warnings case.

Internal plumbing still carries the teardown boolean the individual steps set, including the gate's queued callbacks, and converts once in the public wrapper. That is why the tests injecting a dependency container still drive the internal overload unchanged.

One thing worth a reviewer's opinion: adding a case to a public Swift enum is source-breaking for exhaustive switches in customer code. If we ever want the noisy step named in the result rather than only in the logs, it is much cheaper to decide that before release than after. I left it at two cases rather than guessing at a taxonomy.

Test totals after this change: 841 (723 unit, 15 notification-extension, 103 offline-events), 0 failures.


swift-sdk/SDK/IterableAPI+SwitchProject.swift
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        guard !isRaised else { return false }
+        isRaised = true
+        return true
+    }
+
+    /// Lowers the gate, replays everything queued during the switch in FIFO order against
+    /// the new project, then delivers every registered callback on the main thread.
+    func endSwitch(succeeded: Bool) {
```
rtlsilva: Currently the gate is lowered first, which could potentially allow a new call to overtake calls made during the switch.
For example, a new `setEmail("C")` could run before the previously queued `setEmail("B")`, leaving "B" as the final identity.

Consider keeping the gate raised until all queued calls have finished running.

swift-sdk/SDK/IterableAPI+SwitchProject.swift
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        guard outgoing.apiKey != apiKey else {
+            ITBInfo("switchProject called with the API key already in use. Nothing to tear down.")
+            deliverOnMainThread(callback, true)
+            return
+        }
+
+        // Step 2: raise the gate so any SDK call made from here until step 8 is queued.
+        guard ProjectSwitchGate.shared.beginSwitch(callback: callback) else {
```
rtlsilva: A second `switchProject` call with a different API key is currently ignored, although its callback still runs.
For example, if ProjectA -> ProjectB is running and the app requests ProjectA -> ProjectC, both callbacks fire but the SDK ends on B.

This may be surprising behavior for a multi-region API, but it also seems to be intentional and covered by tests, so this comment might be more of a design discussion than a blocker.

Should we clarify or reconsider the behavior when a second call targets a different API key?
For example, only joining requests targeting the same project and either queueing or rejecting requests targeting a different one?

[Related comment](https://github.com/Iterable/iterable-android-sdk/pull/1086/changes#r3873326598) on the Android PR

swift-sdk/SDK/IterableAPI+SwitchProject.swift
```diff
@@ -0,0 +1,320 @@
...
+        guard !isRaised else { return false }
+        isRaised = true
+        return true
+    }
+
+    /// Lowers the gate, replays everything queued during the switch in FIFO order against
+    /// the new project, then delivers every registered callback on the main thread.
+    func endSwitch(succeeded: Bool) {
+        lock.lock()
+        isRaised = false
```
rtlsilva: Currently the gate is lowered first, which could potentially allow a new call to overtake calls made during the switch.
For example, a new `setEmail("C")` could run before the previously queued `setEmail("B")`, leaving "B" as the final identity.

Consider keeping the gate raised until all queued calls have finished running.

[Related comment](https://github.com/Iterable/iterable-android-sdk/pull/1086/changes#r3873296799) on the Android PR

swift-sdk/Internal/InternalIterableAPI.swift
```diff
@@ -70,6 +70,16 @@ final class InternalIterableAPI: NSObject, PushTrackerProtocol, AuthProvider {
         get {
             localStorage.getAttributionInfo(currentDate: dateProvider.currentDate)
         } set {
+            // Campaign, template and message IDs belong to this instance's project, and the
+            // store they go to is shared with whatever project comes next. Work this instance
+            // owns can resolve after `switchProject` has cleared them — a universal link
+            // redirect returning, or a push tapped while this instance is still installed —
+            // and writing them back would attribute the new project's first event to a
+            // campaign it has never heard of.
+            guard !hasClearedProjectScopedStorage else {
```
rtlsilva, line 86: The flag check and attribution write are not synchronized with `clearProjectScopedStorage`. A late response can pass the check, have the switch clear attribution, and then write the previous project’s attribution back into storage.

Maybe these operations could be serialized, or could asynchronous work carry a project identifier that is checked before saving?

[Related comment](https://github.com/Iterable/iterable-android-sdk/pull/1086/changes#r3873347166) on the Android PR

swift-sdk/SDK/IterableAPI+SwitchProject.swift
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        guard let outgoing = implementation else {
+            ITBError("switchProject called before the SDK was initialized. Initializing with the supplied API key instead.")
+            initialize2(apiKey: apiKey,
+                        launchOptions: nil,
+                        config: config,
+                        apiEndPointOverride: apiEndPointOverride,
+                        dependencyContainer: dependencyContainer,
+                        callback: { started in deliverOnMainThread(callback, started) })
```
rtlsilva: When called before initialization, iOS normally reports `true` if initialization succeeds. Android reports `false`, since no previous project was cleaned up and no device disable was confirmed. The iOS documentation also describes that case as `false`.

This is a parity issue - should this result be aligned across both platforms?

rtlsilva (changes requested):

swift-sdk/SDK/IterableAPI+SwitchProject.swift
rtlsilva: The pending-switch queue fixes A→B followed by C, but A→B followed immediately by A still bypasses it.

While B is in flight, `implementation` remains A, so the active-key guard reports `.switchedCleanly` before the gate can see that B is the in-flight destination. B then lands even though A was requested last.

Suggest consulting the gate before treating the active key as a no-op, and add a rapid A→B→A test.

swift-sdk/SDK/IterableAPI+SwitchProject.swift
```diff
@@ -0,0 +1,442 @@
...
+        guard request.apiKey != inFlightApiKey else {
+            callbacks.append(contentsOf: request.callbacks)
+            return false
+        }
+
+        if let index = pending.firstIndex(where: { $0.apiKey == request.apiKey }) {
+            pending[index].callbacks.append(contentsOf: request.callbacks)
+        } else {
+            pending.append(request)
+        }
```
rtlsilva, line 54: `firstIndex` coalesces with any earlier pending request, which breaks the documented last-request-wins order.
During A→B, requests for C, then D, then C leave `[C, D]` pending; the last C callback fires with the first C switch and the SDK finally settles on D.

Suggest only coalescing a repetition that does not cross another destination, or move the existing request to the tail, and add a B/C/D/C ordering test.

rtlsilva (changes requested):

swift-sdk/SDK/IterableAPI+SwitchProject.swift
rtlsilva: The successful cold-start case is aligned now, but the initialization-failure case still diverges.
Swift forwards `initialize2`'s `started` value, so a failed `start()` reports `.switchedWithWarnings`; Android's pre-initialization branch unconditionally reports `SWITCHED_CLEANLY`.

Since result parity is part of this contract, suggest aligning this edge case and add a failing-start pre-initialization test.

